PR #4, opened 2018-09-19
chore: fix .travis.yml name to match travis needs

cyril-lakech:


cyril-lakech:
@guillaumechervetaxa Le build travis est KC #sorry

PR #81, opened 2019-01-02
fix token expired event

### LINKED ISSUE #77: [context-oidc] Issue when token Expire and SilentRenewFail
When a silent renew faile, the token expire and it is no renewed. So we stuck with a expired token with no posssibility to renew it automatiquely. We have to force logout and login again

youf-olivier:
## Before this PR
After a silent failure, token expired is stuck in his own state. Login is no longer possible

## After this PR
Login  with a expired token possible + silentsignin when token expired

Close #77

### DIFF
--- a/examples/context/package.json
+++ b/examples/context/package.json
@@ -19,5 +19,11 @@
   },
   "scripts": {
     "start": "react-scripts start"
-  }
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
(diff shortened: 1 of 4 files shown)


guillaumechervetaxa:
Yout last comment is not WIP , so it will merge automaticaly if i accept the pr.

Do you want I validate it?

You should configurer well your eslint or prettier plugin, it format ' instead of " and add warning to the linter

PR #151, opened 2019-02-24
English - change all instances of "authentified" to be "authenticated"

citypaul:

Hello again!

This one is an English correction. We wouldn't refer to something as "not authentified", the correct term would be "not authenticated". I've done a search and replace here to replace all instances. It all seems to work fine for me, but please do check it.

I realise this will also change one of the routes, but this is worth it as it's correct English. Hope this is ok!

## Before this PR

Routes and props that were not authenticated were referred to as "not authentified".

## After this PR

These routes and props are using the term "not authenticated".

### DIFF
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -53,7 +53,7 @@ render(<App />, document.getElementById('root'));
 
 ```javascript
 const propTypes = {
-  notAuthentified: PropTypes.node, // react component displayed during authentication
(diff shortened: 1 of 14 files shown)


citypaul:
Thanks @guillaumechervetaxa :)

If I can get a chance this week, I will try to do a few more things. I was thinking about looking at the docs to see if I could help out there a bit if you don't mind?

I will be working on a prototype during the week, so will do what I can when I get a chance. Thanks again for the quick merge!

PR #182, opened 2019-03-27
Fetch Core : Fix for header props

youf-olivier:

### DIFF
--- a/packages/fetch-core/src/fetchToken.js
+++ b/packages/fetch-core/src/fetchToken.js
@@ -11,10 +11,10 @@ export const fetchWithToken = (fetch, getAccessTokenInjected) => async (
 ) => {
   // eslint-disable-next-line
   let headers = new Headers();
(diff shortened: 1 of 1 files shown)


youf-olivier:
This PR will fix an issue repported by @Sebo1000

packages/fetch-core/src/fetchToken.js
```diff
@@ -12,7 +12,7 @@ export const fetchWithToken = (fetch, getAccessTokenInjected) => async (
   // eslint-disable-next-line
   let headers = new Headers();
   // eslint-disable-next-line
-  if (options.headers && !(options.headers instanceof Headers)) {
```
guillaumechervetaxa: This modif will create bugs

I think you want to do this 

 if (options.headers){
  headers =  !(options.headers instanceof Headers) ? new Headers(options.headers) : options.headers;
}

PR #292, opened 2019-09-03
fix typo to get the condition working

Snaptags:
## Before this PR
undefined <= 0 will **always** be false in javascript

## After this PR
if the configuration is empty, the exception gets thrown as expected.


Snaptags:
Sorry, forget to remove the ! at the beginning of the condition.

PR #303, opened 2019-09-23
Add callback component

Madebymaurice:
## Before this PR
The callback component was the only component that we should'nt customize

## After this PR
Now, for the context, or redux package, you can customize by adding a prop

### DIFF
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -23,21 +23,18 @@ The default routes used internally :
 - www.your-app.fr/authentication/not-authorized
 
 ```javascript
-import React from "react";
-import { render } from "react-dom";
-import { BrowserRouter as Router } from "react-router-dom";
-import { AuthenticationProvider, oidcLog } from "@axa-fr/react-oidc-context";
-import Header from "./Layout/Header";
-import Routes from "./Router";
-import oidcConfiguration from "./configuration";
+import React from 'react';
+import { render } from 'react-dom';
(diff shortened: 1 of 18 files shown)


guillaume-chervet:
Awesome PullRequest! Thank you @Madebymaurice 
@youf-olivier what do you think about it?

packages/context/README.md
```diff
@@ -55,6 +52,7 @@ const propTypes = {
   notAuthenticated: PropTypes.node, // react component displayed during authentication
   notAuthorized: PropTypes.node, // react component displayed in case user is not Authorised
   authenticating: PropTypes.node, // react component displayed when about to redirect user to be authenticated
+  callbackComponentOverride: PropTypes.node, // react component displayed when user is connected
```
youf-olivier: I think we should be uniform with the override components names. Your name is pretty good, but if we change the others, we create breaking changes. @guillaume-chervet What do you think ?

youf-olivier:
I would try to understand why we can't use the same logic as the other overrides component. I'll check this afternoon.

Thanks for the PR !

packages/context/README.md
Madebymaurice: You right but as "callback" props is already used I thought it would be a little bit confused to use the same props name. But tell me what you want and I change it ;)

guillaumechervetaxa: Sure, I have few breaking change in mind for 4.0.0.
Let's open some issues about it.

PR #310, opened 2019-10-01
refactor(context) : remove callback OR in core

youf-olivier:


packages/core/src/routes/OidcRoutes.js
```diff
@@ -47,7 +44,7 @@ const OidcRoutes = ({
 
   switch (path) {
     case callbackPath:
-      return <CallbackComponent callbackComponentOverride={callbackComponentOverride} />;
```
guillaumechervetaxa: nice update @youf-olivier ,
you have to update redux package too.

youf-olivier: The problem was in the context package only

guillaume-chervet: May be but thos remove break the feature on redux pacakge :p

youf-olivier: I Don't think so, the redux package didn't use the callbackComponentOverride prop. Or I missed something else ? 

From oidc.js in Redux Package : 

return (
    <OidcProvider store={store} userManager={getUserManager()}>
      <OidcRoutes
        configuration={configuration}
        notAuthenticated={notAuthenticated}
        notAuthorized={notAuthorized}
        callbackComponent={withComponentOverrideProps(
          AuthenticationCallback,
          callbackComponentOverride
        )}
      >
        {children}
      </OidcRoutes>
    </OidcProvider>
  );

guillaumechervetaxa: yep

Issue #317, opened 2019-10-12
[all] F5 always reauthenticate


guillaumechervetaxa:
Bug fixed on version 2.X.X is back in 3.X.X

The correct version to take is
https://github.com/AxaGuilDEv/react-oidc/blob/65bc3a6e0375c1181401eadd8c049bb0710071a4/packages/redux/src/authenticate.js
 
I will fixe it on monday

youf-olivier:
In the same case : when token is expired the silent signin is not launched.

guillaumechervetaxa:
It is not the same case at all Olivier :p

Issue #301, opened 2019-09-18
Redirection to login page, even if user is not active, cause unexpected behaviour


jhujhul:
### Issue and Steps to Reproduce
When a user logs in, after a long period of time (usually between 12h and 24h), he is redirected to a login page, even if he closed his computer session.
In the context of Axa France, I'm talking about this login page:
![Capture](https://user-images.githubusercontent.com/15234610/65162686-eadf3b00-da39-11e9-9db8-72fcf69b458c.PNG)

The issue occurs when he tries to login 10 minutes after the redirection, which happens often when his session is closed: because he took too much time to authenticate, he will be redirected to this page:
![expired](https://user-images.githubusercontent.com/15234610/65161260-8b802b80-da37-11e9-9b8c-8ff6cc67037f.PNG)

### Versions
Package: @axa-fr/react-oidc-redux
Version: 2.0.8

youf-olivier:
I'm not sure that is a real issue, but I already had this case. I will check that in few days. 

Thanks to the repport.

guillaume-chervet:
It is a good use case to add to the library.
I think the best solution is to add a new route with new component that advert we lose the session. Of course this component will be overridable and have by defailt a button for reconnect. We should not make it automaticaly in that case.

PR #319, opened 2019-10-14
fix(all): F5 always re-auth and manage session lost

guillaumechervetaxa:
#317 and #301

### DIFF
--- a/packages/core/src/services/oidcServices.js
+++ b/packages/core/src/services/oidcServices.js
@@ -8,20 +8,20 @@ export const isRequireAuthentication= ({ user, isForce }) =>
 export const isRequireSignin  = (oidcUser, isForce) =>
   isForce || !oidcUser;
 
-export const authenticateUser = (userManager, location, history, user) => async (isForce = false) => {
-  if (!userManager || !userManager.getUser) {
-    return;
-  }
+export const authenticateUser = (userManager, location, history, user=null) => async (isForce = false) => {
   let oidcUser = user;
   if(!oidcUser)
   {
-    await userManager.getUser();
+    oidcUser = await userManager.getUser();
   }
   if(userRequested)
   {
     return;
   }
   const url = location.pathname + (location.search || '');
+  
+  console.log('oidcUser', oidcUser);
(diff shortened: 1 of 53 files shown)


youf-olivier:
Even we are an autosquash, clean commits are nicer to follow you on the PR. (for next time ;))

packages/core/src/services/oidcServices.js
```diff
@@ -1,20 +1,39 @@
...
-  if (isRequireAuthentication(oidcUser, isForce) && !userRequested) {
+  const url = location.pathname + (location.search || '');
+
+  if (isRequireSignin(oidcUser, isForce)) {
     oidcLog.info('authenticate user...');
-    const url = location.pathname + (location.search || '');
     userRequested = true;
     await userManager.signinRedirect({ data: { url } });
     userRequested = false;
+  } else if (oidcUser && oidcUser.expired) {
```
youf-olivier: We never had to do that, don't you think that the problem is deeper ?

guillaume-chervet: Je vais essayer de finir cette pr la semaine prochaine. J'avais presque fini.

Shashank-11:
Any chance of merging this PR?

youf-olivier:
Yes. We have 3 PR to merge. December was a very high intensive month, and for now Guillaume is on vacation. We close all PR start of January

packages/core/src/services/oidcServices.js
guillaumechervetaxa: Oui c'est ce qui uniformise le comportement des 2 librairies.

guillaumechervetaxa:
Ready for review @youf-olivier,

I do not remember the property you used to shorten the accesstoken lifetime for testing.
Do you know which property it is?

youf-olivier:
> Ready for review @youf-olivier,
> 
> I do not remember the property you used to shorten the accesstoken lifetime for testing.
> Do you know which property it is?

It is the client id in config file. With IS4 demo you can set the client Id to spa.short to get a shorlive token.

PR #330, opened 2019-11-04
doc(oidc-metadata): add document about metadata

guillaumechervetaxa:
related to #327


youf-olivier:
Two dots (:) miss for a correct commit message

PR #439, opened 2020-04-24
fix: add logout action

Gparquet:

When performing a disconnection, an authentication was restarted

## Before this PR
An authentication was launched during the disconnection

## After this PR
Additional action allowing a dispatch which launches the disconnection action properly

### DIFF
--- a/packages/context/src/oidcContext/AuthenticationContext.tsx
+++ b/packages/context/src/oidcContext/AuthenticationContext.tsx
@@ -9,6 +9,7 @@ export type oidcContext = {
   events: UserManagerEvents;
   authenticating: ComponentType;
   isLoading: boolean;
+  isLoggingOut: boolean;
   userManager: UserManager;
   error: string;
 };
--- a/packages/context/src/oidcContext/AuthenticationContext.hooks.spec.ts
+++ b/packages/context/src/oidcContext/AuthenticationContext.hooks.spec.ts
@@ -22,6 +22,7 @@ describe('useAuthenticationContextState hook tests suite', () => {
     expect(result.current.oidcState).toEqual({
       error: '',
       isLoading: false,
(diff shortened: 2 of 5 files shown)


packages/context/src/oidcContext/AuthenticationContext.tsx
```diff
@@ -9,6 +9,7 @@ export type oidcContext = {
   events: UserManagerEvents;
   authenticating: ComponentType;
   isLoading: boolean;
+  isLogout: boolean;
```
guillaumechervetaxa, line 12: I think I would call it : isLoginOut

guillaume-chervet: Very Nice :)
Just Waiting for that renaming !

miszo: `isLoggingOut` :)

youf-olivier:
Yes Geoffrey, just change the little thing and go on :)

packages/context/src/oidcContext/AuthenticationContext.tsx
romuleald: `logout` is not a valid english word, `disconnect` would be more accurate, isDisconnecting/isDisconnected would be 100% accurate

miszo:
Hi Guys, do you have any plans releasing new version with this fix?

guillaumechervetaxa:
It was released in the last alpha. But it seem to still have some problem in our application (we are investigating on the subject)

youf-olivier:
@miszo If you can test the last alpha and tell us if you meet some issues it would be very helpfull !

youf-olivier:
Oh it seems there is a problem of contains with last version 

https://github.com/AxaGuilDEv/react-oidc/issues/445

PR #468, opened 2020-06-09
style: remove overwritten property jsx-filename-extension

hamzahamidi:
Eslint remove overwritten property jsx-filename-extension

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,12 +63,6 @@ module.exports = {
     'import/default': error,
     'import/named': error,
     'import/namespace': error,
-    'react/jsx-filename-extension': [
-      warn,
-      {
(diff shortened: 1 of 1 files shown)


youf-olivier:
Can we just change js/jsx by ts/tsx ?

hamzahamidi:
`jsx-filename-extension` was already overwritten in line 94 & only allowed jsx in files with extensions: .jsx & .tsx ( which should be the default setting), as for the files with jsx, this option will raise warnings. 
https://github.com/AxaGuilDEv/react-oidc/blob/dad7d824957ebdcadb5f0c5ba3e8a4b1023a9353/.eslintrc.js#L94

youf-olivier:
@guillaume-chervet

PR #469, opened 2020-06-09
fix(Vanilla): return type callback with signinSilent

hamzahamidi:
The `signinSilentCallback` doesn't return the object 
```javascript
{
    type: "callback"
}
```

### DIFF
--- a/packages/vanilla/src/index.js
+++ b/packages/vanilla/src/index.js
@@ -50,12 +50,11 @@ const init = (config, document = window.document) => {
       .then(() => {
         return { type: "callback" };
       });
-  } else if (
-    _document.location.toString().includes(config.silent_redirect_uri)
-  ) {
-    return _userManager.signinSilentCallback().then(() => {
-      type: "callback";
-    });
+  }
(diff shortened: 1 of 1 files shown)


youf-olivier:
You just removed the else keyword, it doesn't change anything, does it ?

hamzahamidi:
When you want to return an object using arrow functions in JavaScript, you have 2 options:

1. You either use curly braces **but you must use the keyword return**.
2. Or You return the object directly but you need it wrap it with parenthesis.

```javascript
const x = () => {} // return nothing (which is the case in the then of signinSilentCallback)
const y = () => ({}) // returns an object
```

### Before:
The promise returned by signinSilentCallback returned `Promise<void>`

### After:
The promise returned by signinSilentCallback returned Promise wrapping the value  `{type: "callback"}`

youf-olivier:
Yes I missed the ( !!

Good point.

hamzahamidi:
**NB**: Enabling the linter could've prevented this issue. The variable `type` isn't even declared!

youf-olivier:
@guillaume-chervet

PR #473, opened 2020-06-17
fix(context): made type for AuthenticationProviderProps partial

### LINKED ISSUE #472: Not working with typescript [react-oidc-context]
looks like all the builds has been broken with typescript for react-oidc-context.
None of the versions are working with typescript.
Versions: All
Screen Shot 2020-06-16 at 12 51 32 pm

react-oidc-context has a dependency with react-oidc-core. is that version always pointing to the latest?
why are the previous builds failing?Is there a versioning miss-match with `react-oidc-context` and `react-oidc-core` ?

Is there a fall back version?

sgebhardt:
## Before this PR
Typescript throws errors when when a partial configuration is served to the AuthenticationProvider.

## After this PR
This should fix #472
Suggested by @ghostd in https://github.com/AxaGuilDEv/react-oidc/pull/457#issuecomment-644894640_

### DIFF
--- a/packages/context/src/oidcContext/AuthenticationContext.provider.tsx
+++ b/packages/context/src/oidcContext/AuthenticationContext.provider.tsx
@@ -175,7 +175,7 @@ type AuthenticationProviderProps = Omit<AuthenticationProviderIntProps,
   | 'logoutUserInt'
 >
 
-const AuthenticationProvider: ComponentType<AuthenticationProviderProps> = withRouter(
+const AuthenticationProvider: ComponentType<Partial<AuthenticationProviderProps>> = withRouter(
   withServices(AuthenticationProviderInt, {
     CallbackInt: Callback,
     authenticationServiceInt: authenticationService,


ghostd:
You beat me :) I was preparing the PR.

Shashank-11:
@ghostd @sgebhardt hey guys, have you tested this? I did this thing in node modules and strangely, the error still exists for me :(

sgebhardt:
@Shashank-11 yes i have tested it locally in the node_modules folder. 
I added the code into the src file and additionally into the typedefinition file (node_modules\@axa-fr\react-oidc-context\dist\oidcContext\AuthenticationContext.provider.d.ts).

I think the changes in the src file had no effect, because VS Code uses only the code from the dist folder. The compiled JS code runns anyway. It is only the typescript compiler which throws the error.

ghostd:
I did the same test as @sgebhardt, it worked for me.

ghostd:
Hi @guillaumechervetaxa 

Do you know who could add a second approval?

guillaume-chervet:
@oyouf ?

PR #489, opened 2020-07-16
Popup login

Maxinger15:
## Before this PR
Before it was not possible to login via a popup window, even it is supported by oidc-client-js.

## After this PR
I added the support for popups. Use popup_redirect_uri instead of redirect_uri in the config and 
there will open a Popup instead of redirecting you to the login page. With this your SPA keep open during login and no user interactions get lost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ### Changelog
 All notable changes to this project will be documented in this file.
 
+#### [v3.1.4](https://github.com/AxaGuilDEv/react-oidc/compare/v3.1.3...v3.1.4)
+> 14 July 2020
+- fix(context): made type for AuthenticationProviderProps partial [`#473`](https://github.com/AxaGuilDEv/react-oidc/pull/473) by Olivier YOUF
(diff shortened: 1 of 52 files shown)


guillaume-chervet:
Hi @Maxinger15 ,did you manage to make it works?

Why do you need the popup mode? The world is using the default mode ( facebook, google, Microsoft, etc). For the user UX the other mode seems better.

Maxinger15:
Hi, the code itself is running, but I am currently fighting with lerna and the test files because since I changed the files to success the lint test, the spec.ts files throw many errors about false types and interfaces. I am not sure if it is because I changed in manny files type... to interface ... to success the lint. But I am not sure how youbwas able to success the lint in the latest release, becaus I had to edit files i never touched. If you have time next week may we phone and you he

guillaume-chervet:
I'am sorry, I'am out of computer until 17 august. 
@oyouf , do you have time to help @Maxinger15 ?

guillaume-chervet:
Hi @Maxinger15, I really think it is a bad idea to integrate the popup login inside that repository.
I really think it it a bad UX practice and it will couple us more to oidc client. It will also complexify to much the code.

You may publish it from your own fork. @oyouf, @arnaudforaison, what do you think about it?

PR #530, opened 2020-10-06
feat(context-fetch/fetch-core): removing recompose

youf-olivier:
Removing recompose peer dependancie in context-fetch and fetch-core.


packages/fetch-core/src/withFetchRedirectionOn401.tsx
```diff
@@ -0,0 +1,16 @@
+﻿import { wrapAuthenticating } from './withFetchRedirectionOn403';
+
+export const fetchWithRedirectionOn401 = (fetch: typeof window.fetch, history: ReactOidcHistory) => async (
+  url: RequestInfo,
+  options: RequestInit = { method: 'GET' }
+) => {
+  const response = await fetch(url, options);
+  if (response.status === 401) {
+    history.push('/authentication/not-authenticated');
```
romuleald, line 9: **issue:** URL should be configurable, also, it may not force redirection is user do not want to, could generate frustration here

packages/context-fetch/src/withFetchToken.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type WindowFetch = typeof fetch;
+
+const withUser = (WrappedComponent: React.ComponentType) => (props: any) => {
+  const { oidcUser } = useReactOidc();
+  return <WrappedComponent {...props} user={oidcUser} />;
+};
+
+const withTokenFromFetchToken = (fetch: WindowFetch) => (WrappedComponent: React.ComponentType) => (props: any) => {
+  const { fetch: newFetch } = fetchToken(fetch)(props);
```
romuleald, line 15: **suggestion:** `newFetch` could be named something like `fetchWithToken`

packages/context/src/oidcContext/AuthenticationContext.provider.tsx
```diff
@@ -165,15 +165,10 @@ export const AuthenticationProviderInt = ({
...
-  | 'CallbackInt'
-  | 'setLoggerInt'
-  | 'OidcRoutesInt'
-  | 'oidcLogInt'
-  | 'authenticateUserInt'
-  | 'logoutUserInt'
->
+type AuthenticationProviderProps = Omit<
+  AuthenticationProviderIntProps,
+  'authenticationServiceInt' | 'CallbackInt' | 'setLoggerInt' | 'OidcRoutesInt' | 'oidcLogInt' | 'authenticateUserInt' | 'logoutUserInt'
```
romuleald, line 170: **suggestion:** you could use an `enum` here

packages/fetch-core/src/withFetchRedirectionOn403.tsx
```diff
@@ -0,0 +1,26 @@
...
+    history.push('/authentication/not-authorized');
+  }
+  return response;
+};
+
+export const wrapAuthenticating = (fetchWithRedirectioInjected: typeof fetchWithRedirectionOn403) => (fetch: typeof window.fetch) => (
+  WrappedComponent: React.ComponentType
+) => (props: any) => {
+  const history = useHistory();
+  const previousFetch = fetch || props.fetch;
```
romuleald, line 19: **question:** I don't understand what is happening here, the `fetch` you give in param is mandatory and should be a `window.fetch`, why would you need the `||`?

packages/fetch-core/src/withFetchRedirectionOn401.tsx
youf-olivier: You can be create issue about it, indeed

packages/context-fetch/src/withFetchToken.tsx
youf-olivier: Good point

packages/context/src/oidcContext/AuthenticationContext.provider.tsx
youf-olivier: For a Type ?

arnaudforaison (approved):

packages/fetch-core/src/withFetchRedirectionOn403.tsx
youf-olivier: I can't be more agree with you about lisibility, the first fetch is just for the tests.

I think I will work on better code lisibility later.

arnaudforaison (approved):

guillaume-chervet (approved):
Nice pr Olivier :)

xballoy:
Nice job, I was stuck with recompose because of this peer dependency :)

youf-olivier:
Tiens ca pas été mergé ici

xballoy (approved):

xballoy (approved):

PR #590, opened 2021-03-02
Allowing for more complex route paths for built in authentication end…

S1r-Lanzelot:
…points

Example return values from new function:

image

## Before this PR
Before this pr the path of my app, `https://myhost/someapp` would not work with the built in authentication routes

## After this PR
This path, `https://myhost/someapp` will work while maintaining original functionality

Non breaking change that helps make the route logic more accommodating. Let me know if I could make this cleaner... Not sure how fast you all can merge and push if it looks good, coming in on a deadline but would prefer to continue to use this npm from the public registries.

Thanks guys!


S1r-Lanzelot:
If the general idea is acceptable, I will update PR to fix prettier failure. Thanks

S1r-Lanzelot:
This only solves half of the problem, still need to support the router push functionality.

Issue #440, opened 2020-04-24
'No matching state found in storage' when pressing back button after login


bogdbo:
### Issue and Steps to Reproduce
1. `git clone https://github.com/AxaGuilDEv/react-oidc.git`
2. `cd react-oidc/examples/context`
3. `npm install`
4. `npm run start`
5. Open http://localhost:3000 in a browser
6. Press Login
7. Use `bob`/`bob`
8. Once logged in press `back` button once
9. This error shows up (No matching state found in storage)

![image](https://user-images.githubusercontent.com/889997/80229265-15d0e380-8648-11ea-86bd-d6067b47d273.png)

### Versions
3.0.8
#### Expected
Should not show any errors
#### Actual
shows errors
### Additional Details

- Installed packages:

guillaume-chervet:
Hi @bogdbo,
Sorry for the delay.
Did you manage to make it works?

bogdbo:
Hey @guillaume-chervet, sorry for the delay as well. It's been a while and I have switched a few projects since then but as far as I can recall we didn't find a solution but I think we might have switched to pure oidc.

guillaume-chervet:
Thank you @bogdbo , I think it is a good issue to fix

wjureczka:
I have same issue, is there any workaround @guillaume-chervet?

guillaume-chervet:
Thank you for the solution sample, that definitively something to add.
I'am quite busy these days, I cannot tell you when I would be able to checkout this.

wjureczka:
Hi @guillaume-chervet , I have debugged some components in order to fix this issue, but I've got stuck. Maybe you will have few minutes to help. So to start with:

We can replace this line: https://github.com/AxaGuilDEv/react-oidc/blob/master/packages/context/src/Callback/Callback.container.tsx#L9
With this: window.history.replaceState({}, window.document.title, window.location.origin + user.state.url);
and then we are properly deleting routes with `/auth/*` path from oidc-core and landing to `user.state.url`.

However, component `<OidcSecure>` is not updated and does not let its children to render, because `history.push`(ReactOidcHistory) is not called. What can I do now to help to fix bug?

VCuzmin:
Hello! What is the status of this issue? I'm asking that because I'm quite blocked. Thank you a lot!

guillaume-chervet:
@youf-olivier do you have time to look at this one?

guillaume-chervet:
[...]

PR #606, opened 2021-04-12
#440 fix for back button click

mpelesh:
This might be a bit simplistic fix for the issue #440, but it works.
Ideally `signinRedirect()` call should be performed by `oidc-client` JS library whenever callback state is missing in the storage (e.g. this is how it is done in by the Keycloak JS adapter), but this will still require a small change to `react-oidc` package.


mpelesh:
@guillaume-chervet , @youf-olivier , sorry for tagging you here.
Could you please review this PR at your convenience.

packages/context/src/Callback/Callback.container.tsx
```diff
@@ -6,15 +6,20 @@ import withServices from '../withServices';
...
   } else {
     oidcLogInternal.warn('no location in state');
   }
 };
 
-export const onRedirectError = (history: ReactOidcHistory, oidcLogInternal: typeof oidcLog) => ({ message }: { message: string }) => {
+export const onRedirectError = (history: ReactOidcHistory, oidcLogInternal: typeof oidcLog, userManager: UserManager) => ({ message }: { message: string }) => {
   oidcLogInternal.error(`There was an error handling the token callback: ${message}`);
-  history.push(`/authentication/not-authenticated?message=${encodeURIComponent(message)}`);
+  if (message === "No matching state found in storage") {
```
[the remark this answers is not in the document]
mpelesh: This is exactly what is causing this problem.
 `oidc-client-js` library saves state in the browser state store before it redirects to OIDC provider. This state is removed by `oidc-client-js` immediately after login redirect. When user presses back button, the browser sends the state key (stored in the browser history) in the request. `oidc-client-js` tries to retrieve the state from the storage using the provided key and fails: https://github.com/IdentityModel/oidc-client-js/blob/a8a70ce68bd07373abd7d33e016a97adc5d204bc/src/OidcClient.js#L120. So when user navigates between different pages using back button, he/she will get an error page instead of going to the previous page.
I am not sure if `oidc-client-js` can throw any other exception in this case, but I did not want to assume this is the only one, so I added this check.
Sorry for the lengthy explanation or if I misunderstood your question.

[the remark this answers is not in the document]
guillaume-chervet: if (message === "No matching state found in storage") { // I really think you should remove this line

I think the subject is a little bit more intricate, we have to define exactly the expected behavior.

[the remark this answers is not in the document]
mpelesh: Made the request change, even though I am not  very comfortable with this.

> I think the subject is a little bit more intricate, we have to define exactly the expected behavior.

Agree. My PR addressing the issue the same way it is implemented in the Keycloak JS adapter: they send login redirect call in case signin redirect callback fails to locate saved state in a browser state store.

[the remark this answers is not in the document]
guillaume-chervet: Thank you @mpelesh, I will make many test with many providers.

packages/context/src/Callback/Callback.container.tsx
```diff
@@ -6,15 +6,15 @@ import withServices from '../withServices';
...
   } else {
     oidcLogInternal.warn('no location in state');
   }
 };
 
-export const onRedirectError = (history: ReactOidcHistory, oidcLogInternal: typeof oidcLog) => ({ message }: { message: string }) => {
+export const onRedirectError = (oidcLogInternal: typeof oidcLog, userManager: UserManager) => ({ message }: { message: string }) => {
   oidcLogInternal.error(`There was an error handling the token callback: ${message}`);
-  history.push(`/authentication/not-authenticated?message=${encodeURIComponent(message)}`);
+  userManager.signinRedirect({ data: { url: "/" } })
```
guillaume-chervet, line 17: I have made some tests it works pretty well.
But you cannot set a default behavior on error. It won't work with all use cases.

mpelesh: Exactly. This is why I had this line to handle missing state only:
```javascript
if (message === "No matching state found in storage") {
```
I don't think there is another way to detect that state is missing in the state store.
Do you want to me revert to the previous version of the fix?

mpelesh: @guillaume-chervet, do you have any feedback on this?

guillaume-chervet: Sorry again for the delay.

I cannot accept the pull request with that line. I think you might remove your change in onRedirectError.
I think you cannot set up this default behavior for every use case. 
You can set it up on your own application by using this event.

What do you think ?

guillaume-chervet: What do you think about this : #627

guillaume-chervet:
The back button problem is fixed in v4.
https://www.npmjs.com/package/@axa-fr/react-oidc-context/v/4.0.0-beta.9

guillaume-chervet:
I Close the PullRequest, 
This is fixed in V4 and V5. 

Thank you very much for your help. The new version is easier to make evolved.

PR #709, opened 2022-03-03
Add button to collapse NavBar

ddecrulle:
## Make the navBar responsive

It's a very little contribution, but in case of debugging with dev tools, in my laptop the navbar is not displayed due to the mid weight width.

## Before this PR
image

## After this PR

 Capture d’écran 2022-03-03 à 09 09 21
Capture d’écran 2022-03-03 à 09 09 13

## Additionnal comment

As you can see in the PR, we don't have the same eslint and prettier config. I think to avoid some useless diff in the commit it could be very usefull to add config inside the project throught a `.eslintrc` and `.prettierrc` file.

Then, in macOS `set PORT=4200` isn't working, what do you think about add [cross env](https://www.npmjs.com/package/cross-env) in dev dependencies ? 

Thanks a lot !

### DIFF
--- a/packages/context/src/App.tsx
+++ b/packages/context/src/App.tsx
@@ -1,20 +1,23 @@
 import React from 'react';
 import { BrowserRouter as Router, Route, Link } from 'react-router-dom';
 import { Home } from "./Home";
-import {Profile, SecureProfile} from "./Profile";
-import {configurationAuth0, configurationIdentityServer} from './configurations';
-import {withOidcSecure,OidcProvider} from "./oidc";
-import {FetchUser} from "./FetchUser";
-import {MultiAuthContainer} from "./MultiAuth";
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you very much @ddecrulle 

Sure add .eslintrc and .prettierrc is a must.

For your question "Then, in macOS set PORT=4200 isn't working, what do you think about add [cross env](https://www.npmjs.com/package/cross-env) in dev dependencies ?"
Yes, I think it is a must. Thank you for your help and feedback !

guillaume-chervet:
It seem. I have to add some condition for the Continuous Integration.

May I merge your PullRequest now?

guillaume-chervet (approved):

ddecrulle:
Thanks for you fast reply ! Of course you can merge right know.

PR #711, opened 2022-03-03
fix(react): Export fetch with token and add Cross Env

### LINKED ISSUE #710: fetchWithToken not exported by the lib
### Is any reason that `fetchWithToken` is not exported by the lib ?

The lib only expose the wrapper `withOidcFetch`. I know that in case of ServiceWorker use, the methode `fetchWithToken` isn't necessary to be exposed. 

But with Firefox it could be usefull. I tried to make my own and to pass token myself but I observe a request sent without token because it's `null` (not yet updated by the lib). 

In the lib code source, I see `export const fetchWithToken` I suspect an export oversight in the index. If there is a raison I'm intrusting to discuss about :)

### Versions
4.0.X

#### Expected

`fetchWithToken` exported 

#### Actual

`fetchWithToken` not exported

ddecrulle:
Fixes #710

Cross env as mentioned [here](https://github.com/AxaGuilDEv/react-oidc/pull/709#issuecomment-1057840799)


guillaume-chervet (approved):
Thank you very much, that perfect

PR #716, opened 2022-03-06
[fix] Make the keepAlive call relative to origin

TeckTn:
Make the keepAlive call relative to origin instead  being relative to the current Page.

## Before this PR
The url https://mydomain.com/page1/section1 will result in a call to https://mydomain.com/page1/OidcKeepAliveServiceWorker.json
## After this PR
The url https://mydomain.com/page1/section1 will result in a call to https://mydomain.com/OidcKeepAliveServiceWorker.json


guillaume-chervet:
Thank You very much @TeckTn !

PR #731, opened 2022-03-17
fix(session): session lost on refresh

### LINKED ISSUE #730: [Question] How do I persist user state (e.g. refreshing page) without service worker?

Before anything, huge thanks for the library, it is amazing and you should all be proud of it and how well you are maintaining it!

I have a question regarding persisting user state (i.e. user is still logged in) after refreshing a page. I've used your [demo](https://black-rock-0dc6b0d03.1.azurestaticapps.net/profile) and changed it locally to test. I only commented the configuration below:

```
    client_id: 'interactive.public.short', //  interactive.public.short
    redirect_uri: window.location.origin+'/authentication/callback', // http://localhost:4200/authentication/callback
    scope: 'openid profile email api offline_access',
    authority: 'https://demo.identityserver.io',
    refresh_time_before_tokens_expiration_in_second: 70,
    //service_worker_relative_url:'/OidcServiceWorker.js',                  // commented here
    //service_worker_only: false,                                           // commented here
```

With this, the demo should use the memory storage, which in fact works (the access token is not hidden by the service worker). 

image

However, when I refresh the page, the user is no longer authenticated 🤔 

image

Keep in mind, if I had used the service worker, **it would have worked and persisted the user state between refreshes**. However, when I choose not to, it does not.

I was under the impression the access and ID tokens were saved in local storage, that way this doesn't happen. Am I missing something? How can I achieve normal behaviour (refreshing and user still be logged in) without using a service worker?

Thank you!

### Versions

Latest version

#### Expected
Expected user to not be asked to login again after refreshing page.

#### Actual
User loses "logged in" status and has to login again.

### Additional Details
None

guillaume-chervet:
fix #730

### DIFF
--- a/MIGRATION_GUIDE_V3_TO_V4.md
+++ b/MIGRATION_GUIDE_V3_TO_V4.md
@@ -89,6 +89,7 @@ const propTypes = {
     refresh_time_before_tokens_expiration_in_second: PropTypes.number,
     service_worker_relative_url: PropTypes.string,
     service_worker_only: PropTypes.boolean, // default false
+    extras: StringMap|undefined // ex: {'prompt': 'consent', 'access_type': 'offline'} list of key/value that are send to the oidc server (more info: https://github.com/openid/AppAuth-JS)
   }).isRequired
 };
 ```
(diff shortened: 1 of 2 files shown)


LuchoTurtle:
Thanks a lot @guillaume-chervet ! You're a life-saver 🙏

PR #743, opened 2022-03-31
refactor(hook): Decoupling user authenticated flag on hook call

LuchoTurtle:
Hey there! Love this library, it's really neat, huge props! 🎉 

This PR attempts to decouple the information from the `useOidcUser()` hook. I am aware that I can use the `withOidcSecure()` hook to secure a component to check if the user is authenticated or not and redirect to the login page. However, the latter is not useful when I'm dealing with multiple libraries for different providers (e.g., I use the [`msal-react`](https://www.npmjs.com/package/@azure/msal-react) library for Azure sign-in alongside this one for other on-prem identity providers like IdentityServer or CAS Apereo).

I'm decoupling because `useOidcUser()` causes unnecessary re-renders when I only want to know if the user is authenticated or not. This greatly  impacts the performance of my application currently, especially if it's a root component.

#### Current hook

The following picture shows me rendering a new component on the demo project and showing the amount of times it renders when I'm calling:

```
    const isAuthenticated = useOidcUser().isLogged;     // -> the old one
    console.log("render")

    return (
        <div className={isAuthenticated ? "card text-white bg-success mb-3" : "card text-white bg-warning mb-3"}>
            <div className="card-body">
                <h5 className="card-title">{isAuthenticated ? "User is authenticated" : "User is not authenticated"}</h5>
            </div>
        </div>
    )
```

image

#### New hook
With the new hook I added - `useUserIsAuthenticated()` - it only renders once 😄 .

```
    const isAuthenticated = useUserIsAuthenticated();    // -> the new one
    console.log("render")

    return (
        <div className={isAuthenticated ? "card text-white bg-success mb-3" : "card text-white bg-warning mb-3"}>
            <div className="card-body">
                <h5 className="card-title">{isAuthenticated ? "User is authenticated" : "User is not authenticated"}</h5>
            </div>
        </div>
    )
```

image

This logic of decoupling hooks can further be extended to the rest of the components of `useOidcUser()`, even though I guess this one is more alarming, in my opinion. This is useful so I don't have to use libraries like [`react-hooks-in-callback`](https://github.com/fernandoem88/react-hooks-in-callback) to control re-renders.

### DIFF
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -299,6 +299,30 @@ const DisplayUserInfo = () => {
 };
 ```
 
+## How to get check if user is authenticated : Hook Method
+
+You can use the `useUserIsAuthenticated()` hook to check if the user is logged-in/authenticated. You can also use the `useOidcUser()` hook as well but this one is less noisy - it does not cause unnecessary re-renders.
+
+```javascript
+import { useUserIsAuthenticated } from '@axa-fr/react-oidc-context';
+
+const DisplayIdToken =() => {
+  const isLoggedIn = useUserIsAuthenticated();
+
+  if(!isLoggedIn){
+    return <p>you are not authentified</p>
+  }
+
+  return (
+          <div className="card text-white bg-info mb-3">
+            <div className="card-body">
+              <h5 className="card-title">User is authenticated! :)</h5>
+            </div>
+          </div>
+  );
+}
+```
+
 # Service Worker Support
 
 - Firefox : tested on firefox 98.0.2
--- a/packages/context/README.md
+++ b/packages/context/README.md
@@ -158,7 +158,7 @@ import {useOidc} from "./oidc";
 
 export const Home = () => {
 
-    const { login, logout, isLogged} = useOidc();
+    const { login, logout, isAuthenticated} = useOidc();
     
     return (
         <div className="container-fluid mt-3">
@@ -166,8 +166,8 @@ export const Home = () => {
                 <div className="card-body">
                     <h5 className="card-title">Welcome !!!</h5>
                     <p className="card-text">React Demo Application protected by OpenId Connect</p>
-                    {!isLogged && <button type="button" className="btn btn-primary" onClick={() => login('/profile')}>Login</button>}
-                    {isLogged && <button type="button" className="btn btn-primary" onClick={logout}>logout</button>}
+                    {!isAuthenticated && <button type="button" className="btn btn-primary" onClick={() => login('/profile')}>Login</button>}
+                    {isAuthenticated && <button type="button" className="btn btn-primary" onClick={logout}>logout</button>}
                 </div>
             </div>
         </div>
@@ -176,7 +176,7 @@ export const Home = () => {
 
 ```
 The Hook method exposes : 
-- isLogged : is the user logged?
+- isAuthenticated : if the user is logged in or not
 - logout: logout function (return a promise)
 - login: login function 'return a promise'
 
@@ -194,7 +194,7 @@ const AdminSecure = () => (
   </OidcSecure>
 );
 
-// adding the oidc user in the props
+// adding the oidc user in the propsis
 export default AdminSecure;
 ```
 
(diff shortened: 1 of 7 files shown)


guillaume-chervet:
hi @LuchoTurtle thank you very much for the PullRequest!

May be all isLogged should be renamed as isAuthenticated. It is a breaking change but I really prefer your namming 'isAuthenticated"

I do think, you missed the "useOidc" existing hook, that will not cause multiple rerendering (see code bellow). May be adding a new hook is not necessary. What do you think ?
```
export const useOidc =(configurationName=defaultConfigurationName) =>{
    const getOidc =  Oidc.get;
   
    const login = (callbackPath=undefined) => {
        return getOidc(configurationName).loginAsync(callbackPath);
    };
    const logout = () => {
        return getOidc(configurationName).logoutAsync();
    };
    
    let isLogged = false;
    const oidc = getOidc(configurationName);
    if(oidc){
        isLogged = getOidc(configurationName).tokens != null;
    }
    
    return {login, logout, isLogged };
}
```

LuchoTurtle:
Hey @guillaume-chervet!

Thanks for the fast reply, you're amazing pal!
 
> May be adding a new hook is not necessary. What do you think ?
You're totally right. I quite forgot about that hook and that I could use the property. It runs smoothly, no problems there, thank you!

> and may be useOidcUser hook sould return another kind of data for not confusing people. like
> 
> {user, userLoadingStatus} = useOidcUser()
> 
> and status be like :
> const loadingStatus:{
> "notAuthenticated",
> "loading",
> "loaded"
> }

I agree, no reason on repeating the same information when one is clearly more performant than the other. With this in mind, I committed the changes you request and also optimized the `useOidcUser()` hook, saving one unnecessary render along the way 😄 .


> May be all isLogged should be renamed as isAuthenticated. It is a breaking change but I really prefer your namming 'isAuthenticated"
I'm with you on this one! It's how MSAL uses it and I wager it's more intuitive for the folk around. Also added this *breaking change* to the commit so if you happen to merge it, you oughta 5.0 the library!

packages/context/src/oidc/User.ts
```diff
@@ -1,31 +1,37 @@
...
+            setOidcUser({...oidcUser, status: UserStatus.Loading});
+            oidc.userInfoAsync()
                 .then((info) => {
                     if (isMounted) {
-                        setOidcUser(info);
-                        setIsOidcUserLoading(false);
+                        setOidcUser({user: info, status: UserStatus.Loaded});
                     }
                 })
+                .catch(() => setOidcUser({...oidcUser, status: UserStatus.Unauthenticated}));
```
guillaume-chervet, line 30: Thank you again for your PullRequest !

May be add an "LoadingError" state like

export enum UserStatus {
    Unauthenticated= 'Unauthenticated',
    Loading = 'Loading user',
    Loaded = 'User loaded',
    LoadingError = "Loading error"
}

What do you think?

packages/context/src/oidc/User.ts
```diff
@@ -1,31 +1,37 @@
...
-    if(oidc){
-        isLogged = oidc.tokens != null;
-    }
-    
-    return {oidcUser, isOidcUserLoading, isLogged: isLogged}
-}
\ No newline at end of file
+    }, []);
+
+    return {oidcUser: oidcUser.user, isOidcUserLoading: oidcUser.status}
```
guillaume-chervet, line 36: May be rename that property
isOidcUserLoading => oidcUserLoadingState 

What do you think ?

packages/context/README.md
```diff
@@ -194,7 +194,7 @@ const AdminSecure = () => (
   </OidcSecure>
 );
 
-// adding the oidc user in the props
+// adding the oidc user in the propsis
```
guillaume-chervet, line 197: may by a typo here?

packages/context/src/oidc/User.ts
LuchoTurtle: Makes sense! Going to change it.

packages/context/src/oidc/User.ts
LuchoTurtle: I was trying to make it the same as the old one to reduce compatibility concerns but since we already have breaking changes, I do agree it's more intuitive!

packages/context/README.md
LuchoTurtle: Oops! 😅

guillaume-chervet:
Thank you very much @LuchoTurtle , it will be breaking. It will generate a new 5.0.0 version. I will make a migration guide in another PullRequest.
I prefer make a breaking now before developers switch from v3 to v4.

Issue #779, opened 2022-05-07
Callback does not redirect back to the requested url


masterbater:
### Issue and Steps to Reproduce

 Requesting to url  /profile if unauthenticated it will go to the oidc provider to initiate login, then it will redirect to callback ,  but during testing it didnt redirect to profile its just stuck to  Authentication complete You will be redirected to your application.  Is there a config to set to redirect after the callback? Do you store the requested url like localstorage or cookie then use that to redirect after callback ?
### Versions
     5.8.0-alpha0
     
 Config 
```
 export const configuration = {
  client_id: "testingapp-auth",
  redirect_uri: window.location.origin + "/authentication/callback",
  silent_redirect_uri:
    window.location.origin +
    "/authentication/callback/authentication/silent-callback", // Optional activate silent-signin that use cookies between OIDC server and client javascript to restore the session
  scope: "openid profile email offline_access roles",
  authority: "http://oidcprovider.com/apps/testingapp",
    service_worker_relative_url: "/OidcServiceWorker.js",
    service_worker_only: true,
};

```

masterbater:
Nevermind, it is customizable I can provide my own component thanks. I found it here. 
https://github.com/AxaGuilDEv/react-oidc/blob/master/packages/react/README.md

guillaume-chervet:
Hi @masterbater,  thank you very much for the issue?
You should not be stuck to that page. It should redirect you to your /profile page.
It should be a bug or something like that 😜 
Do you have an error message in the console or something else ?

masterbater:
I just created my own CallbackComponent, I supplied useEffect with idToken and redirect to desired path if its not null.

guillaume-chervet:
May you copy Paste your work solution here?
It will help me to set better defaut component.

Thodor12:
@masterbater I'm currently running into the same issue, could you share your solution?

masterbater:
> @masterbater I'm currently running into the same issue, could you share your solution?

```
import { useOidcIdToken } from "@axa-fr/react-oidc";
import React, { useEffect } from "react";
import { useNavigate } from "react-router-dom-v5-compat";

const CallbackSuccessComponent = () => {
  const navigate = useNavigate();
  const { idToken, idTokenPayload } = useOidcIdToken();
  useEffect(() => {
    if (idToken) {
      navigate("/", { replace: true });
      window.location.reload();
    }
  }, [idToken]);

  return <div>CallbackSuccessComponent</div>;
[...]

PR #780, opened 2022-05-08
fix(callbakc): Callback does not redirect back to the requested url

guillaume-chervet:
#779

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -56,7 +56,7 @@ jobs:
         
       - id: publish-react
         uses: JS-DevTools/npm-publish@v1
-        if: github.ref == 'refs/heads/master'
+        #if: github.ref == 'refs/heads/master'
         with:
           token: ${{ secrets.NPM_TOKEN }}
           package: ./packages/react/package.json
@@ -77,7 +77,7 @@ jobs:
 
       - id: publish-react-context
         uses: JS-DevTools/npm-publish@v1
-        if: github.ref == 'refs/heads/master'
+        #if: github.ref == 'refs/heads/master'
         with:
           token: ${{ secrets.NPM_TOKEN }}
           package: ./packages/context/package.json
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -56,7 +56,7 @@ jobs:
         
       - id: publish-react
         uses: JS-DevTools/npm-publish@v1
-        #if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         with:
           token: ${{ secrets.NPM_TOKEN }}
           package: ./packages/react/package.json
@@ -77,7 +77,7 @@ jobs:
 
       - id: publish-react-context
         uses: JS-DevTools/npm-publish@v1
-        #if: github.ref == 'refs/heads/master'
+        if: github.ref == 'refs/heads/master'
         with:
           token: ${{ secrets.NPM_TOKEN }}
           package: ./packages/context/package.json
--- a/packages/context/package.json
+++ b/packages/context/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@axa-fr/react-oidc-context",
-  "version": "5.7.5",
+  "version": "5.7.6-alph0",
   "private": false,
   "main": "dist/index.js",
   "jsnext:main": "dist/index.js",
--- a/packages/context/package.json
+++ b/packages/context/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@axa-fr/react-oidc-context",
-  "version": "5.7.6-alph0",
+  "version": "5.7.6-alpha1",
   "private": false,
   "main": "dist/index.js",
   "jsnext:main": "dist/index.js",
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -501,3 +501,43 @@ More information about OIDC
 - Chrome/Edge : tested on version upper to 90
 - Opera : tested on version upper to 80
 - Safari : tested on Safari/605.1.15
+
+# FAQ
+
+## Do not set OidcProvider Between Router and its routes
+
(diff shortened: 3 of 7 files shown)


Thodor12 (commented):
I'm not sure if this tackles the root cause, this seems to only affect cases where there is no state provided, which (provided in the original issue linked) isn't the case. I tested this locally just now by changing up the code to this and it doesn't fix it.

Thodor12:
Isn't it more likely that `isMounted` becomes false because you're not awaiting the promise?

guillaume-chervet:
Thank you for your help. I cannot reproduce your error.
I am interrested why do you have isMouted=false

Can you share more of you code logic ?

Thodor12:
> Thank you for your help. I cannot reproduce your error. I am interrested why do you have isMouted=false
> 
> Can you share more of you code logic ? @Thodor12

My code logic isn't anything amazingly spectacular, I'm only using the OidcSecure component inside my router to lock everything down after the /dashboard route, I am however not using a silent signin URL (because of the iframe restrictions due to X-Frame-Options header).

Full config:
```js
function getAuthority() {
    if (Environment.OIDC_AUTHORITY !== undefined) {
        return Environment.OIDC_AUTHORITY;
    } else {
        return `${window.location.protocol}//auth.${window.location.host}`;
    }
}

export function GetOIDCConfiguration(): OidcConfiguration {
    return {
        client_id: Environment.OIDC_CLIENT_ID!,
        authority: getAuthority(),
        redirect_uri: window.location.origin + "/authentication/callback",
        scope: "openid profile email offline_access",
        service_worker_relative_url: '/OidcServiceWorker.js',
        service_worker_only: true
    }
};
```

Thodor12:
However my best guess is still that [on this line](https://github.com/AxaGuilDEv/react-oidc/blob/c9aca5e82e2e450b6a31e04176455768b2c12ced/packages/react/src/oidc/core/default-component/Callback.component.tsx#L41) you're just calling the async method, which is essentially just a promise and not executed (and awaited).
This then causes the return function to be called first, which according to React it's documentation is a cleanup function which will reset `isMounted` to false, after this happens the promise will start to execute and will end up in the first if statement seeing that `isMounted` is false.
It's not a straight guarantee that this might happen, it depends on the device it's thread scheduling system what is executed first, which is probably why it's inconsistent.

Thodor12:
I'll try the alpha0 version you just published after I'm done with work 👍🏻

Thodor12:
The issue still exists in alpha0

guillaume-chervet:
Hi @Thodor12 ,
can you share some code with your provider and more?

isMounted should never be called. You should have a behavior that refresh your OidcProvider.

Thodor12:
In any case here is the router, maybe that is enough info?
```tsx
export function Router() {
    return (
        <BrowserRouter>
            <OidcProvider configuration={GetOIDCConfiguration()}>
                <Routes>
                    <Route element={<Layout />}>
                        <Route path="/" element={<Home />} />

                        <Route path="/dashboard" element={<OidcSecure><Outlet /></OidcSecure>}>
                            <Route index element={<Dashboard />} />
                            <Route path="worklists">
                                <Route index element={<WorkListIndex />} />
                            </Route>
                        </Route>
                    </Route>
                </Routes>
            </OidcProvider>
        </BrowserRouter>
    )
}
```

Thodor12:
Also I can confirm that `isMounted` isn't the issue, I just tested by placing some loggers inside the dist JS file manually to see what the variable is, and it was true. So I think it does push an entry into the history, but like you said there's nothing which refreshes the OidcProvider.

guillaume-chervet:
What happen if you set the code like that?

`export function Router() {
    return (
 
            <OidcProvider configuration={GetOIDCConfiguration()}>
       <BrowserRouter>
                <Routes>
                    <Route element={<Layout />}>
                        <Route path="/" element={<Home />} />

                        <Route path="/dashboard" element={<OidcSecure><Outlet /></OidcSecure>}>
                            <Route index element={<Dashboard />} />
                            <Route path="worklists">
                                <Route index element={<WorkListIndex />} />
                            </Route>
                        </Route>
                    </Route>
                </Routes>
        </BrowserRouter>
            </OidcProvider>

    )
}`

Thodor12:
> What happen if you set the code like that?
> 
> ` export function Router() { return (
> 
> ```
>         <OidcProvider configuration={GetOIDCConfiguration()}>
>    <BrowserRouter>
>             <Routes>
>                 <Route element={<Layout />}>
>                     <Route path="/" element={<Home />} />
> 
>                     <Route path="/dashboard" element={<OidcSecure><Outlet /></OidcSecure>}>
>                         <Route index element={<Dashboard />} />
>                         <Route path="worklists">
>                             <Route index element={<WorkListIndex />} />
>                         </Route>
>                     </Route>
>                 </Route>
>             </Routes>
>     </BrowserRouter>
>         </OidcProvider>
> 
> )
> ```
> 
> } `
> 
> note: I tried it and i can reproduce your error. I will specify it inside the readme.

Unfortunately this one still didn't work for me

Thodor12:
I'm gonna try isolate the problem by removing some of my other dependencies, maybe they're the issue

Thodor12:
I've finally found the determining factor, it works with the setup you've described, provided you do NOT use `<React.StrictMode>`.
For some reason if you wrap your application within this component (which is what CRA does by default) it blocks the redirect from happening, something that you could look into?

guillaume-chervet:
Thank you yeah definitly it is necessary to investigate

jgregory-apogee:
I was able to work around this for now by wrapping my `<OidcProvider>` tags outside of `<React.StrictMode>`.  That lets me keep strict mode for the rest of my codebase.

guillaume-chervet:
Which version of react are you using?
I do not have this problem on mine wicth strict mode with react 17

guillaume-chervet:
Oh thank you i can reproduce it. It will be easier to fix !

PR #789, opened 2022-05-30
feat(oidc): NextJS  Compatibility

tbogdal:
## Before this PR

Could not use **react-oidc** in ***NextJS*** based application, because ***Server Side Rendering*** (aka ***SSR***) would fail on `undefined` reference to `window`.

Additionally, on redirect back to the ***NextJS*** application, the final redirect would not work properly, the router would not detect the URL/History changes. 

## After this PR

In NodeJS the window object does not exists, the global object is the closes alternative. When **react-oidc** is executed on the server side (in Node JS), it uses `global` object conditionally instead of `window`.

When returning back from the IDP, the default URL/History modification is not detected by ***NextJS*** router. The `CallbackManager` has been extended to allow an externally defined custom history object witch implements the `replaceState` method. By extending the `CallbackManager` in this way, we can use the native NextJS router to handle URL/History changes by providing a custom implementation.

Changes introduced:

 * In `Timer.js` when `window` is **undefined**, `global` object is used instead (prevents SSR errors)
 * Extended interface, to allow passing custom history object via `withCustomHistory` prop

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -156,6 +156,7 @@ const propTypes = {
     service_worker_relative_url: PropTypes.string,
     service_worker_only: PropTypes.boolean, // default false
     extras: StringMap|undefined // ex: {'prompt': 'consent', 'access_type': 'offline'} list of key/value that are send to the oidc server (more info: https://github.com/openid/AppAuth-JS)
+    withCustomHistory: PropTypes.function, // Override history modification, return instance with replaceState(url, stateHistory) implemented (like History.replaceState()) 
   }).isRequired
 };
 ```
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -155,7 +155,7 @@ const propTypes = {
     refresh_time_before_tokens_expiration_in_second: PropTypes.number,
     service_worker_relative_url: PropTypes.string,
     service_worker_only: PropTypes.boolean, // default false
-    extras: StringMap|undefined // ex: {'prompt': 'consent', 'access_type': 'offline'} list of key/value that are send to the oidc server (more info: https://github.com/openid/AppAuth-JS)
+    extras: StringMap|undefined, // ex: {'prompt': 'consent', 'access_type': 'offline'} list of key/value that are send to the oidc server (more info: https://github.com/openid/AppAuth-JS)
     withCustomHistory: PropTypes.function, // Override history modification, return instance with replaceState(url, stateHistory) implemented (like History.replaceState()) 
   }).isRequired
 };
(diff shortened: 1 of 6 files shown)


packages/react/src/oidc/core/routes/OidcRoutes.tsx
```diff
@@ -22,6 +23,7 @@ type OidcRoutesProps = {
   configurationName:string;
   redirect_uri: string;
   silent_redirect_uri?: string;
+  withCustomHistory?: () => CustomHistory;
```
guillaume-chervet, line 26: Awesome thank you for your PullRequest,

Do you have a sample of your implementation of that fonction in NextJS ?

packages/react/README.md
```diff
@@ -156,6 +156,7 @@ const propTypes = {
     service_worker_relative_url: PropTypes.string,
     service_worker_only: PropTypes.boolean, // default false
     extras: StringMap|undefined // ex: {'prompt': 'consent', 'access_type': 'offline'} list of key/value that are send to the oidc server (more info: https://github.com/openid/AppAuth-JS)
```
jeremyrambaud, line 158: Missing a comma

packages/react/src/oidc/core/routes/OidcRoutes.tsx
tbogdal: Yes,

Here is an excerpt, it's very simple:


```typescript
const MyApp: React.FC<AppProps> = ({ Component, pageProps: { session, ...pageProps }, router }) => {
  const [loading, setLoading] = useState(router.asPath.includes('access_token'));

  const store = useStore(pageProps.initialReduxState);
  let searchRedirectPage: PageUrl;

  const withCustomHistory: () => CustomHistory = () => {
    return {
      replaceState: (url?: string | null, stateHistory?: WindowHistoryState): void => {
        router.replace({
          pathname: url,
        });
      },
    };
  };
  // Code omitted...

  return !loading ? (
    <>
      <Head>
        <meta
          name="viewport"
          content="width=device-width, height=device-height,  initial-scale=1.0, user-scalable=no;user-scalable=0;"
        />
      </Head>
      <OidcProvider configuration={OIDC_CONFIGURATION} withCustomHistory={withCustomHistory}>
        <Provider store={store}>
          <RouterScrollProvider>{layout}</RouterScrollProvider>
        </Provider>
      </OidcProvider>
    </>
  ) : null;
};

```

packages/react/README.md
tbogdal: Will fix the typo....

packages/react/src/oidc/core/routes/OidcRoutes.tsx
guillaume-chervet: this would be awesome to add more docmentation and a demo after that PullRequest.

Thank you very much !

guillaume-chervet (approved):

guillaume-chervet:
I need you to resolve the conflit in order to merge the PullRequest. Woulb be awesome!

PR #792, opened 2022-06-02
feat(logout_uri): add post logout dynamic url for external authenticat…

LucassvTI:
## After this PR
Added post logout dynamic url for external authenticator redirection and corrected query separator parameters of ? for &

### DIFF
--- a/packages/react/src/oidc/vanilla/oidc.ts
+++ b/packages/react/src/oidc/vanilla/oidc.ts
@@ -628,21 +628,17 @@ Please checkout that you are using OIDC hook inside a <OidcProvider configuratio
          timer.clearTimeout(this.timeoutId);
      }
      
-    async logoutAsync(callbackPath: string | undefined = undefined, extras:StringMap=null, postLogoutRedirectUri: string | undefined = undefined) {
+    async logoutAsync(callbackPathOrUrl: string | undefined = undefined, extras: StringMap = null) {
         const configuration = this.configuration;
         const oidcServerConfiguration = await this.initAsync(configuration.authority, configuration.authority_configuration);
-        if(callbackPath && (typeof callbackPath !== 'string'))
+        if(callbackPathOrUrl && (typeof callbackPathOrUrl !== 'string'))
         {
-            callbackPath = undefined;
-            console.warn('callbackPath path is not a string');
+            callbackPathOrUrl = undefined;
+            console.warn('callbackPathOrUrl path is not a string');
         }
-		if(postLogoutRedirectUri && (typeof postLogoutRedirectUri !== 'string'))
-        {
-            postLogoutRedirectUri = undefined;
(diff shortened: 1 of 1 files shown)


packages/react/src/oidc/vanilla/oidc.ts
```diff
@@ -628,24 +628,29 @@ Please checkout that you are using OIDC hook inside a <OidcProvider configuratio
          timer.clearTimeout(this.timeoutId);
      }
      
-    async logoutAsync(callbackPath: string | undefined = undefined, extras:StringMap=null) {
+    async logoutAsync(callbackPath: string | undefined = undefined, extras:StringMap=null, postLogoutRedirectUri: string | undefined = undefined) {
```
guillaume-chervet, line 631: I would prefer something like that.

`
 async logoutAsync(callbackPathOrUrl: string | undefined = undefined, extras:StringMap=null) {
        const configuration = this.configuration;
        const oidcServerConfiguration = await this.initAsync(configuration.authority, configuration.authority_configuration);
        if(callbackPath && (typeof callbackPath !== 'string'))
        {
            callbackPath = undefined;
            console.warn('callbackPath path is not a string');
        }
        const path = (callbackPath === null || callbackPath === undefined) ? location.pathname + (location.search || '') + (location.hash || '') : callbackPath;
		const isUri = callbackPathOrUrl.includes("https://") || callbackPathOrUrl.includes("http://");
		let url = isUri ? callbackPathOrUrl : window.location.origin + path;
        // @ts-ignore
        const idToken = this.tokens ? this.tokens.idToken : "";
        await this.destroyAsync();  
        if(oidcServerConfiguration.endSessionEndpoint) {
            let extraQueryString = "";
            if(extras){
                for (let [key, value] of Object.entries(extras)) {
                    extraQueryString +=`&${key}=${encodeURIComponent(value)}`;
                }
            }
            window.location.href = `${oidcServerConfiguration.endSessionEndpoint}?post_logout_redirect_uri=${encodeURIComponent(url)}&id_token_hint=${idToken}${extraQueryString}`;
}
`

guillaume-chervet:
thank you very much @LucassvTI this is awesome.
I made you a feeback

packages/react/src/oidc/vanilla/oidc.ts
guillaume-chervet: I prefer that solution. I found i cleaner and simpler to use

LucassvTI: I like this solution.

guillaume-chervet (approved):

packages/react/src/oidc/vanilla/oidc.ts
guillaume-chervet: Thank you very much @LucassvTI. 
Awesome!

Issue #741, opened 2022-03-30
source maps


pinggi:
### Issue and Steps to Reproduce

The source maps are enabled, but Typescript files are missing in the `node_modules/@axa-fr/react-oidc-context/dist` folder. The browser is unable to load the file when debugging an error/exception.

### Versions

"@axa-fr/react-oidc-context": "^4.5.8" (4.5.8 in yarn.lock)

#### Expected

The Typescript files are there or source maps are disabled.

guillaume-chervet:
hi @pinggi ,

Thank you for your issue. WIth version of create-react-app are you using?
I have some source map available in my dist folder when I build the library:

![image](https://user-images.githubusercontent.com/52236059/160934366-4008527a-515e-46d5-bbf3-c74d5b56988f.png)

pinggi:
I use version 5.0.0.

1. I installed the package: `yarn add @axa-fr/react-oidc-context`
2. I noticed there are missing *.ts files because the browser I use for debugging couldn't load the *.ts files

If I open `node_modules/@axa-fr/react-oidc-context/dist/vanilla/oidc.d.ts.map` file for example, I can see the source path leads to `../../src/oidc/vanilla/oidc.ts` which resolves to `node_modules/@axa-fr/react-oidc-context/src/oidc/vanilla/oidc.ts` - I don't have the file there (the `src` folder is missing there).

guillaume-chervet:
Hi, i understand the problem. I will investigate.

guillaume-chervet:
I just push a fix to include src folder to npm publish.
Thank you very much for your issue !

guillaume-chervet:
I close the PullRequest @pinggi , feel free to reopen it if you need it.
Thank you very much for your help !

PR #825, opened 2022-07-21
Add source code to context package

johngrimes:
Source maps don't really work if the files that they refer to are not present in the package.

This change brings the context package in line with the react package, which correctly includes the `src` directory.

This was originally addressed as part of #741, but the change seems to have been reversed in 403f28d0e1ad121bc4fbf7f76ee381c98a120d8f.

### DIFF
--- a/packages/context/package.json
+++ b/packages/context/package.json
@@ -11,6 +11,7 @@
   },
   "files": [
     "dist",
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
thank you very much @johngrimes

PR #828, opened 2022-08-03
fix(oidcprovider): remove semicolon from error output

DASPRiD:
error

## Before this PR
Semicolon would occur after login errors in output.

## After this PR
Semicolon will not occur after login errors in output.


guillaume-chervet:
Thank you very much @DASPRiD :)

PR #841, opened 2022-08-23
Replace Link with NavLink to show the current Page

radu-cs:
## Before this PR
image

## After this PR
image


guillaume-chervet:
Thank you very much @radu-cs :)

PR #858, opened 2022-09-14
basic cookie pattern

kainanftf:
Hello @guillaume-chervet, thanks again for the library!

Currently I have a niche situation where I am forced to use cookies in order to store JWTs, to support the view layer of an older RoR application(which has additional CSRF protections) alongside newer React/Next apps. I don't want to get too much into the details about that, but rest assured the tokens will still be sent through authorization header for any AJAX requests. Sadly, I cannot use the service worker so cookies are probably about === to localStorage in the XSS department, but with the added benefit of enabling a single location for the JWTs for both the legacy app and newer Next/React ones.

Anyways, I think this code/branch probably would not be helpful to other consumers of your library, but I was hoping to get some personal advice if you have the spare time. I was hoping you could take a look and let me know if theres any glaring problems that I would need to address, or if there is some tests I could potentially run? 

Also, as you probably have noticed, theres a bit of weirdness with the code, which I hope to explain below:

Originally, I wanted to simply wrap the setters/getters for `storage[] =`/`storage[]`, to keep the main logic inside the `initSession` object the same. I would then be able to update the set/gets to optionally use cookies instead of a client-side store. 
Something like:
```js
const set = (key, value, useCookie) => {
  if (useCookie) // do cookie logic and return
  // otherwise do the usual
  return storage[key] = value
}

// and then do something like
const setTokens () => {
  // old
  storage[`oidc.${configurationName}:${redirectUri}`] = JSON.stringify({tokens});
  // new
  helper.set(`oidc.${configurationName}:${redirectUri}`, JSON.stringify({tokens});

}
```
However, I quickly ran into issues with fitting the whole object inside a single cookie, plus I wanted to be able to specify the cookie names. I found that because of all the cookie splitting/merging, the basic set/get abstraction wasn't preventing additional logic from being added into the `initAsync` object methods(although in hindsight, perhaps I could move the splitting/merging of cookies into more complex helper methods in the helper object instead)...Anyway the cookie size limitation, is the main reason for the splitting and merging.

Also I was experiencing some issues getting logout to function properly.
If I don't delete the all the cookies during the `clearAsync()`, I am unable to logout the user. Removing the cookies during the`clearAsync()`  function seems to remedy the problem. It also seems that when deleting the "main cookie"(the one that has the accessToken/status key) it triggers the `clearAsync()` method which clears the remaining the cookies(maybe? not sure if thats whats actually happening yet). 

On the surface, it also appears that the splitting of the tokens object creates an odd situation where pieces of the object are enough to refresh the other bits:
https://user-images.githubusercontent.com/85503587/190058831-77736da7-8830-4ada-ba43-70bbfd180a43.mp4 

Lastly, even after deleting the cookies I seem to get a quick screen flash before the unauthenticated popup unless I don't set the initial cookie in the `initAsync` method: 
https://user-images.githubusercontent.com/85503587/190058915-4c9a027d-abe1-4b0b-b6d0-5f59bd4ef9c0.mp4
I was wondering if there is any potential implications/issues that could arise from not doing that bit?

"Working Example":
https://user-images.githubusercontent.com/85503587/190059137-74c6014c-06ba-4a03-8bdf-4b685dc58afc.mp4

So, this is a lot and I'm sure you're quite busy...if you don't have time to look over this I totally understand and am still very thankful for the library...However, if you do find the time any feedback/assistance is appreciated!

Best,
Kainan


guillaume-chervet:
hi @kainanftf ,

Thank you for your pull request, i understand your need.

You do not need to modify react-oidc, you can implement you own Storage implementation it would be simpler.

`
/** This Web Storage API interface provides access to a particular domain's session or local storage. It allows, for example, the addition, modification, or deletion of stored data items. */
interface Storage {
    /** Returns the number of key/value pairs. */
    readonly length: number;
    /**
     * Removes all key/value pairs, if there are any.
     *
     * Dispatches a storage event on Window objects holding an equivalent Storage object.
     */
    clear(): void;
    /** Returns the current value associated with the given key, or null if the given key does not exist. */
    getItem(key: string): string | null;
    /** Returns the name of the nth key, or null if n is greater than or equal to the number of key/value pairs. */
    key(index: number): string | null;
    /**
     * Removes the key/value pair with the given key, if a key/value pair with the given key exists.
     *
     * Dispatches a storage event on Window objects holding an equivalent Storage object.
     */
    removeItem(key: string): void;
    /**
     * Sets the value of the pair identified by key to value, creating a new key/value pair if none existed for key previously.
     *
     * Throws a "QuotaExceededError" DOMException exception if the new value couldn't be set. (Setting could fail if, e.g., the user has disabled storage for the site, or if the quota has been exceeded.)
     *
     * Dispatches a storage event on Window objects holding an equivalent Storage object.
     */
    setItem(key: string, value: string): void;
    [name: string]: any;
}`


So you just need to set

`
var youOwnCookieStorage = {}; // Your implementation

export const configurationIdentityServer = {
    client_id: 'interactive.public.short',
    redirect_uri: window.location.origin+'/authentication/callback',
    silent_redirect_uri: window.location.origin+'/authentication/silent-callback',
    //silent_login_uri: window.location.origin+'/authentication/silent-login',
    scope: 'openid profile email api offline_access',
    authority: 'https://demo.duendesoftware.com',
    //authority_time_cache_wellknowurl_in_second: 60* 60,
    refresh_time_before_tokens_expiration_in_second: 10,
    service_worker_relative_url:'/OidcServiceWorker.js',
    service_worker_only: false, 
    storage: youOwnCookieStorage,      // <------------------------------------  Set your own storage
    //silent_login_timeout: 3333000
    //monitor_session: true,
  };


`

kainanftf:
Ahh, thank you so much! I will go ahead and do this approach instead! Appreciate the quick response!

kainanftf:
@guillaume-chervet 
Hey wanted to say thanks and share the results of your advice! I used a [proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) so that I could use custom bracket notation setters/getters(not sure if theres a better way to do that). Here's the first draft:

```ts
// Helper used by handler
const initCookieHelper = () => {
  const set = (name: string, value: any) => {
    document.cookie = name + '=' + value + '; path=/';
  };

  const get = (name: string) => {
    const nameEQ = name + '=';
    const ca = document.cookie.split(';');

    for (let i = 0; i < ca.length; i++) {
      let c = ca[i];
      while (c.charAt(0) == ' ') c = c.substring(1, c.length);
      if (c.indexOf(nameEQ) == 0) {
        return c.substring(nameEQ.length, c.length);
      }
    }
    return null;
  };

  const erase = (name: string) => {
    document.cookie =
      name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
  };
  return { set, erase, get };
};

// The CookieHandler Initializer
const initCookieHandler = (
  tokenKey: string,
  identifier: string = 'test'
) => {
  const cookieHelper = initCookieHelper();

  const tokenPaths = [
    `${identifier}.accessToken`,
    `${identifier}.idToken`,
    `${identifier}.refreshToken`,
    `${identifier}.rest`,
  ];

  const clearSplitCookies = () => {
    tokenPaths.forEach((tp) => cookieHelper.erase(tp));
  };

  const setSplitCookies = (tokens: { [key: string]: any }) => {
    const {
      accessToken,
      idToken,
      refreshToken,
      accessTokenPayload,
      expiresIn,
      expiresAt,
      idTokenPayload,
      issuedAt,
      tokenType,
      scope,
    }: {
      [key: string]: string | number | any;
    } = { ...tokens };
    cookieHelper.erase(identifier);
    cookieHelper.set(`${identifier}.accessToken`, accessToken);
    cookieHelper.set(`${identifier}.idToken`, idToken);
    cookieHelper.set(`${identifier}.refreshToken`, refreshToken);
    cookieHelper.set(
      `${identifier}.rest`,
      JSON.stringify({
        accessTokenPayload,
        expiresIn,
        expiresAt,
        idTokenPayload,
        issuedAt,
        tokenType,
        scope,
      })
    );
  };

  const get = (_target: any, key: string) => {
    if (key !== tokenKey) return cookieHelper.get(key);

    const tokenCookies = tokenPaths
      .map((tp) => cookieHelper.get(tp))
      .filter(Boolean); // strips falsey matches

    // if pieces are missing clear em all
    if (tokenCookies.length < 4) {
      clearSplitCookies();
    } else {
      return JSON.stringify({
        tokens: {
          accessToken: cookieHelper.get(tokenPaths[0]),
          idToken: cookieHelper.get(tokenPaths[1]),
          refreshToken: cookieHelper.get(tokenPaths[2]),
          ...JSON.parse(cookieHelper.get(`${identifier}.rest`)!),
        },
      });
    }

    return cookieHelper.get(identifier);
  };

  const set = (_obj: any, prop: string, value: string) => {
    if (prop !== tokenKey) {
      cookieHelper.set(prop, value);
      return true;
    }

    const tokens = JSON.parse(value).tokens;

    if (!tokens) {
      clearSplitCookies();
      cookieHelper.set(identifier, value);
      return true;
    }

    const keys = Object.keys(tokens);
    if (
      keys.includes('accessToken') &&
      keys.includes('accessTokenPayload')
    ) {
      setSplitCookies(tokens);
      return true;
    }

    clearSplitCookies();
    cookieHelper.set(identifier, value);
    return true;
  };

  return {
    set,
    get,
  };
};

// exporting the Proxy Object/CookieStore
export const cookieStore = (tokenKey: string, identifier: string) =>
  new Proxy({}, initCookieHandler(tokenKey, identifier));
```
And I can pass in into the config like:

```ts
const configuration {
  storage: cookieStore(
    'oidc.default:http://localhost:3000/home', // this needs to be same as the path that is used to store tokens in initSessions.ts
    'test' // this is the tokens prefix
  ),
}
```
It seems to be working(although we shall see over time):

https://user-images.githubusercontent.com/85503587/190235179-7a5d6806-0c55-427b-a5c7-5eb09abb5eb0.mp4



So yeah, like you said, we don't have to modify/fork the lib and we can keep getting those updates :D ! 
But seriously thank you again, you're a life saver!

kainanftf:
Thank you, as long as it's not broken you have my permission for sure!

kainanftf:
Hey just a quick update for the snippet, added a expiry argument for the cookies setter, so it can be synced up with with token lifespans:

```ts
// Helper used by handler
const initCookieHelper = () => {
  const set = (
    name: string,
    value: any,
    expiry: number | string = 10
  ) => {
    document.cookie =
      name + '=' + value + ';' + 'max-age=' + expiry + ';' + 'path=/';
  };

  const get = (name: string) => {
    const nameEQ = name + '=';
    const ca = document.cookie.split(';');

    for (let i = 0; i < ca.length; i++) {
      let c = ca[i];
      while (c.charAt(0) == ' ') c = c.substring(1, c.length);
      if (c.indexOf(nameEQ) == 0) {
        return c.substring(nameEQ.length, c.length);
      }
    }
    return null;
  };

  const erase = (name: string) => {
    document.cookie =
      name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
  };
  return { set, erase, get };
};

// The CookieHandler Initializer
const initCookieHandler = (
  tokenKey: string,
  identifier: string = 'test'
) => {
  const cookieHelper = initCookieHelper();

  const tokenPaths = [
    `${identifier}.accessToken`,
    `${identifier}.idToken`,
    `${identifier}.refreshToken`,
    `${identifier}.rest`,
  ];

  const clearSplitCookies = () => {
    tokenPaths.forEach((tp) => cookieHelper.erase(tp));
  };

  const setSplitCookies = (tokens: { [key: string]: any }) => {
    const {
      accessToken,
      idToken,
      refreshToken,
      accessTokenPayload,
      expiresIn,
      expiresAt,
      idTokenPayload,
      issuedAt,
      tokenType,
      scope,
    }: {
      [key: string]: string | number | any;
    } = { ...tokens };
    cookieHelper.erase(identifier);
    cookieHelper.set(
      `${identifier}.accessToken`,
      accessToken,
      expiresIn
    );
    cookieHelper.set(`${identifier}.idToken`, idToken, expiresIn);
    cookieHelper.set(
      `${identifier}.refreshToken`,
      refreshToken,
      3600 // PLEASE SET THIS TO REFRESH TOKEN LIFE SPAN
    );
    cookieHelper.set(
      `${identifier}.rest`,
      JSON.stringify({
        accessTokenPayload,
        expiresIn,
        expiresAt,
        idTokenPayload,
        issuedAt,
        tokenType,
        scope,
      }),
      expiresIn
    );
  };

  const get = (_target: any, key: string) => {
    if (key !== tokenKey) return cookieHelper.get(key);

    const tokenCookies = tokenPaths
      .map((tp) => cookieHelper.get(tp))
      .filter(Boolean); // strips falsey matches

    // if pieces are missing clear em all
    if (tokenCookies.length < 4) {
      clearSplitCookies();
    } else {
      return JSON.stringify({
        tokens: {
          accessToken: cookieHelper.get(tokenPaths[0]),
          idToken: cookieHelper.get(tokenPaths[1]),
          refreshToken: cookieHelper.get(tokenPaths[2]),
          ...JSON.parse(cookieHelper.get(`${identifier}.rest`)!),
        },
      });
    }

    return cookieHelper.get(identifier);
  };

  const set = (_obj: any, prop: string, value: string) => {
    if (prop !== tokenKey) {
      cookieHelper.set(prop, value);
      return true;
    }

    const tokens = JSON.parse(value).tokens;

    if (!tokens) {
      clearSplitCookies();
      cookieHelper.set(identifier, value);
      return true;
    }

    const keys = Object.keys(tokens);
    if (
      keys.includes('accessToken') &&
      keys.includes('accessTokenPayload')
    ) {
      setSplitCookies(tokens);
      return true;
    }

    clearSplitCookies();
    cookieHelper.set(identifier, value);
    return true;
  };

  return {
    set,
    get,
  };
};

// exporting the Proxy Object/CookieStore
export const cookieStore = (tokenKey: string, identifier: string) =>
  new Proxy({}, initCookieHandler(tokenKey, identifier));

```

PR #866, opened 2022-09-16
Allow service worker to accept any domain

### LINKED ISSUE #865: Option to support "globally allowed" OidcTrustedDomains.js
OidcTrustedDomains.js is currently forced to be filled in and the service worker will then only function on these domains.
I've made a tenant like system where multiple domains end up in the exact same application and it's impossible for me to know which application urls these are (they come from the database).

It would be nice that in certain scenarios that you could say
```ts
const trustedDomains = {
    default: ["*"]
};  
```
To allow any domain to use the service worker.

The way I currently intend to solve this is by making a small REST api endpoint which simulates the content of OidcTrustedDomains.js which takes priority over the application itself but I don't know how (or if) that is gonna work.

This has a little bit of haste behind it as I cannot currently get my deployed application to work without this feature.

Thodor12:
closes #865 

## Before this PR
The service worker would always validate trusted domains, it doesn't have support for a wildcard.

## After this PR
The service worker can now accept a wildcard trusted domain to automatically accept any domain (this includes the hasToSendTokens check aswell as the checkDomain checks)

### DIFF
--- a/packages/react/src/oidc/vanilla/OidcServiceWorker.js
+++ b/packages/react/src/oidc/vanilla/OidcServiceWorker.js
@@ -2,6 +2,8 @@
 
 const id = Math.round(new Date().getTime() / 1000).toString();
 
+const acceptAnyDomainToken = "*";
+
 const keepAliveJsonFilename = "OidcKeepAliveServiceWorker.json";
 const handleInstall = (event) => {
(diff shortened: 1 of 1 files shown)


Thodor12:
@guillaume-chervet Build failure because missing token in CI

guillaume-chervet (approved):

PR #872, opened 2022-09-27
fix(type): Expose fetch type from "useOidcFetch" correctly

Thodor12:
Explanation

## Before this PR
The type of the fetch object that `useOidFetch` exposes does not match the type signature of the global `fetch` function, which makes it a bit difficult to make wrapper hooks (or other things) on top of this fetch instance.

## After this PR
By using typescript to extract the parameters of the global `fetch` type and passing them back into `fetchWithToken` it keeps the method signature identical, after that we can use the spread syntax to expand the parameters back into the original requested arguments.


guillaume-chervet:
Thank you very much again @Thodor12  that awesome.

PR #890, opened 2022-10-22
task(config): Set jsx type to "react-jsx".  Remove npm i task in build

jafin:
- set jsx type to `react-jsx` (https://stackoverflow.com/questions/67776707/ts-config-jsx-reactjsx-vs-react)
- remove `npm i` call from package.json `build`, should not be required.


guillaume-chervet (approved):
Thank you @jafin. Your 2 pullrequest are awesome!

PR #891, opened 2022-10-25
feat(eslint): Add ESLint, fix most reported lint error/warnings, fix a bug and add better type

jafin:
This is a large changeset, mainly due to autofix formatting, feel free to decline it or propose a preferred PR submit size.

I grabbed a default set of ESLint rules, if there are any rules that are not aligned to the project, let me know and I'll fix them up (such as single vs double quotes etc)

# Summary of changes.

# ESLINT
- Autofix formatting
- Add ESLINT Config
- adjust boolean variables to naming convention  `'prefix': ['is', 'with', 'should', 'has', 'can', 'did', 'will']`
- remove empty return in `SilentLogin`
- avoid using `object.hasOwnProperty`, prefer `Object.prototype.hasOwnProperty`
- replace `map` with `forEach` (as map did not return value)
- avoid aliasing `this` with self, use `=>`
- `===` preferred over `==`

# Types
- Improve typescript defs
- Remove `Function` ban type, prefer alternatives.
- Export type `OidConfiguration`

# Build
- Reduce `dist` for `react-context` to `src/oidc` and `src/override` (from `src/` which included the demo app)
- Set version of peer deps to React 17/18 instead of `x`

# Further investigation
- Some of the lint rules were commented inline to avoid changing intent of code
- No changes to useEffect hook dependencies, (6 eslint problems reported)


guillaume-chervet (approved):
Thank you so much @jafin , this is just a huge PullRequest. 
You also fixed some types and one bug!
Big Thank you :)

PR #905, opened 2022-11-16
fix(errorMessages): Error message to show correct filename

jafin:
Error message to show correct filename `OidcTrustedDomains.js` instead of `TrustedDomains.js`
Aligns to actual filename in repository.

## Before this PR

thrown Error would show

```text
is not trusted, please add domain in TrustedDomains.js
```

## After this PR

```text
is not trusted, please add domain in OidcTrustedDomains.js
```


guillaume-chervet (approved):
Thank you @jafin

PR #938, opened 2023-01-14
fix: remove global regex and use without flags

abhilashlr:

"JavaScript [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) objects are stateful when they have the [global](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) or [sticky](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) flags"

Since the g flag was used in regex, it caused matches to fail everytime the regex was run. If your client had a higher time set for refreshing access token, the first request would fail and the rest would succeed. And if the time to refresh access token is shorter, all the requests would fail with a 401.

Reason: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

### DIFF
--- a/packages/react/src/oidc/vanilla/OidcServiceWorker.js
+++ b/packages/react/src/oidc/vanilla/OidcServiceWorker.js
@@ -217,7 +217,7 @@ const getCurrentDatabaseDomain = (database, url) => {
                 let domain = domainsToSendTokens[i];
 
                 if (typeof domain === 'string') {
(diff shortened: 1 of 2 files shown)


guillaume-chervet:
Thank you very much @abhilashlr .

PR #958, opened 2023-02-09
fix(oidc): wrong parameter order for internalFetch

TheChristophe:
other functions (e.g. performRevocationRequestAsync) were calling internalFetch and passing timeoutMs as 3rd parameter, but it was taken as numberRetry

## Before this PR

`internalFetch` was defined as
```ts
const internalFetch = async (url, headers, numberRetry = 0, timeoutMs = 10000) => {
...
```

functions like `performTokenRequestAsync` were calling internalFetch like 
```ts
    const response = await internalFetch(url, {
        method: 'POST',
        headers: {
            'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
        },
        body: formBodyString,
    }, timeoutMs);
    // ^ 3rd param is numberRetry
```

## After this PR

swapped `internalFetch` parameters to get
```ts
const internalFetch = async (url, headers, timeoutMs = 10000, numberRetry = 0) => {
...
```
now timeoutMs is in the right place

### DIFF
--- a/packages/react/src/oidc/vanilla/requests.ts
+++ b/packages/react/src/oidc/vanilla/requests.ts
@@ -26,7 +26,7 @@ export const fetchFromIssuer = async (openIdIssuerUrl: string, timeCacheSecond =
     return new OidcAuthorizationServiceConfiguration(result);
 };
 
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you very much @TheChristophe. It is awesome.

PR #967, opened 2023-02-23
fix(oidc): Reduce Service Worker keep-alive response time to allow responses to succeed

dcopestake:

I hope this is okay to submit and I apologise if I've missed something here;

We've been having a lot of issues with failing Service Worker keep-alive requests being logged to our third-party logging platform since implementing `react-oidc` for auth. Hopefully it isn't intentional that these _should_ be erroring out.

## Before this PR

In Chrome (and possibly other browsers) Service Worker keep-alive requests to `/OidcKeepAliveServiceWorker.json` would result in an errored fetch as the browser was timing out at the [maximum of 5 minutes](https://source.chromium.org/chromium/chromium/src/+/main:net/socket/client_socket_pool.cc;l=36) for a request.

## After this PR

Service Worker keep-alive requests return successfully with a 200 and the expected response of `{}` in between 2.5 and 5 minutes before starting the next keep-alive request.


NickCain:
Hi All,

I am also experiencing issues around the keep alive with the timeout error. This looks like it could help me with my issue as well.

guillaume-chervet:
hi @dcopestake , thank you very much for the feedback+ pullrequest.
It may be a little bit more complexe.
The time must be adjusted in that file to keep service worker alive.
https://github.com/AxaGuilDEv/react-oidc/blob/cfd4da5704a8911e6790a3919c2e05bf08be3bf4/packages/react/src/oidc/vanilla/initWorker.ts#L115
And the five minutes may be still a good idea during login phase. It keep ServiceWorker alive 5 minutes on android when we are not on the web Application/service worker domain during login phase.

dcopestake:
Ah okay, I think I understand @guillaume-chervet, thanks. I'll see if I can amend this to take that into account.

dcopestake:
Hi @guillaume-chervet. Apologies for the delay in getting back on this one. I've added 29162f540f15700c464451ee3339d7193b37df38 which I hope addresses your concerns. The minimum delay for the keep alive request is now passed from the client to the service worker.

I haven't specifically changed the delay depending on the phase (i.e. before logging in vs after logging in etc) because unfortunately _any_ request that takes longer than 5 minutes is going to fail in Chrome, so instead I've used 240 seconds as the minimum delay for Android and 150 seconds otherwise.

dcopestake:
Apologies @guillaume-chervet, didn't realise the package had moved on since I originally raised this. Merged in the upstream now so everything should be up to date.

guillaume-chervet (approved):
Thank you again @dcopestake 
let's try it !

PR #970, opened 2023-02-28
fix(oidc): Add missing groups attribute

alex198710:
When adding groups as scope to oidc settings file, the OidcUserInfo is missing this attribute


guillaume-chervet:
Thank you @alex198710

PR #975, opened 2023-03-01
refactor(types): Improve types in OidcServiceWorker. 

jafin:
Improve the types in OidcSerivceWorker

@guillaume-chervet  Could you please look at the below lines.  I've had to ts-ignore two of the lines because of invalid return types. Is what is there correct or needs review?

The forced type assertion may just need a review of the types not to be null.

```js
// forced type assertion, could be null?
200: const { isValid, reason } = isTokensOidcValid(tokens, nonce, currentDatabaseElement.oidcServerConfiguration as OidcServerConfiguration); 
399: return fetchPromise.then(hideTokens(currentDatabase as OidcConfig));

//expects promise, but is void response
343: event.waitUntil(event.respondWith(fetch(newRequest)));
444: event.waitUntil(event.respondWith(maPromesse));
```

can these be written as
```ts
event.waitUntil(fetch(newRequest));
event.waitUntil(maPromesse);
```

EDIT: Ok i see that just `event.WaitUntil` seems to break things , respondWith `appears ok` 
```ts
event.respondWith(fetch(newRequest));
event.respondWith(maPromesse);
```

or is there some special purpose to the code?

### DIFF
--- a/packages/react/service_worker/.eslintrc.js
+++ b/packages/react/service_worker/.eslintrc.js
@@ -1,0 +1,18 @@
+module.exports = {
+  extends: [__dirname + '/config/defaultEslintConfig'],
+  parserOptions: {
+    project: '../tsconfig.eslint.json',
+    tsconfigRootDir: __dirname,
+  },
+  rules: {
+    '@typescript-eslint/naming-convention': [
+      'error',
+      {
+        selector: 'variable',
+        types: ['boolean'],
+        format: ['PascalCase'],
+        prefix: ['is', 'with', 'should', 'has', 'can', 'did', 'will'],
+      },
+    ],
+  },
+};
--- a/packages/react/service_worker/OidcServiceWorker.ts
+++ b/packages/react/service_worker/OidcServiceWorker.ts
@@ -1,5 +1,100 @@
+type Domain = string | RegExp;
+
+type TrustedDomains = {
+    [key: string]: Domain[]
+}
+type OidcServerConfiguration = {
+    revocationEndpoint: string;
+    issuer: string;
+    authorizationEndpoint: string;
+    tokenEndpoint: string;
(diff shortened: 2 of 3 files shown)


guillaume-chervet:
Thank you agin a lot @jafin ,

Has I remember, 
//expects promise, but is void response
343: event.waitUntil(event.respondWith(fetch(newRequest)));
444: event.waitUntil(event.respondWith(maPromesse));

Are like that because of firefox. I have lost a lot of time to make it work on firefox whitout any debug mode available.

jafin:
Thanks @guillaume-chervet 

As is PR is for types, I will not make any change to logic and left the ts-ignore markers in.
There is one logic change , the addition of `if (headers.has(key))` check :

```ts
const serializeHeaders = (headers: Headers) => {
    const headersObj: Record<string,string> = {};
    for (const key of (headers as FetchHeaders).keys()) {
        if (headers.has(key)) {
            headersObj[key] = headers.get(key) as string;
        }
    }
```
`headers.get(key)` can return `null`

```ts
const serializeHeaders = (headers: Headers) => {
    const headersObj: Record<string,string> = {};
    for (const key of (headers as FetchHeaders).keys()) {
        if (headers.has(key)) {
            headersObj[key] = headers.get(key) as string;
        }
    }
```

### Justification

headers in fetch has type HeadersInit which cannot accept type with null value.

```ts
type HeadersInit = [string, string][] | Record<string, string> | Headers;
```

```ts
fetch(originalRequest, {
                        body: newBody,
                        method: clonedRequest.method,
                        headers: {
```

guillaume-chervet:
Really thank you @jafin for your help. I'am agree with your change.

Do you want to merge your PullRequest?

jafin:
@guillaume-chervet yes if you are happy with the change, thanks.

PR #977, opened 2023-03-02
docs(oidctrusteddomains): Add note about configuring wildcards for trusteddomains

### LINKED ISSUE #865: Option to support "globally allowed" OidcTrustedDomains.js

jafin:
Add note about configuring wildcards for `oidctrusteddomains.js` and caution against using them.

- minor linting to the readme.

Should resolve #865

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,10 +1,10 @@
 # @axa-fr/react-oidc
 
 [![Continuous Integration](https://github.com/AxaGuilDEv/react-oidc/actions/workflows/npm-publish.yml/badge.svg)](https://github.com/AxaGuilDEv/react-oidc/actions/workflows/npm-publish.yml)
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you again  @jafin .

It can also be a real warning in logs. if only "*" pattern is set.

guillaume-chervet (approved):

jafin:
Good point, I missed that entirely and assumed regex was only way to do this, should of read the code.. Will put up another PR.

PR #981, opened 2023-03-04
feat(build): Build OidcServiceWorker through esbuild/vite.

jafin:
## Proposal

This is a proposal to run `OidcServiceWorker.ts` through esbuild (Vite) to achieve
- Testability against non service-worker specific code.
- Size reduced file size of oidcServiceWorker (~11kb vs ~23.3kb)

Optionally distribute map and ts src files for debugging.

### Cons

- If a consumer is accustomed to changing OidcServiceWorker.js directly, they may need to amend the source ts, and compile from source instead of directly editing the minified js file.  (Only a minor inconvenience, and unsure what percentage of user would actually need to amend the worker file.)

What I've included is a start on this which has:
- Build/test with Vite/vitetest
- Copy map/js/ts files to public folder
- Test coverage for *some* utility code. Note: I am uncertain how to test service-worker specific code atm.

@guillaume-chervet If this is useful, I'll continue and extract non service-worker specific code into a module for testing, and add basic test coverage.

## Changes

`type: module` was added to `packages.json` to ensure vite build will output js not mjs file.  This then had onflow effect eslint needs to be defined as cjs.

> If the package.json does not contain "type": "module", Vite will generate different file extensions for Node.js compatibility. .js will become .mjs and .cjs will become .js.


guillaume-chervet:
Hi @jafin thank you so much.
Sure you can continue to extract the code to test it. 
It is my aim to clean the hole code and test it to prevent regression and to continu confident evolution.
I'am lacking of time but I'am trying to do it from small step.

Thank you so much for your help.

jafin:
@guillaume-chervet  PR should be ready to review now.

#### Signature change:

```ts
const getCurrentDatabaseDomain = (
  database: Database,
  url: string,
  trustedDomains: TrustedDomains //added
) 
```

* Code that remains in `oidcServerWorker` is serviceworker/eventListener handlers.  Other has been refactored out and tests added.
* typedef d.ts and .map files copied to dist

guillaume-chervet:
Awesome @jafin , i test it today. I will try to give feedback as quickly as i can.

guillaume-chervet:
I cannot look at it today. Tomorrow morning your pullrequest is my priority @jafin .

packages/react/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@axa-fr/react-oidc",
   "version": "6.15.8",
   "private": false,
+  "type":"module",
```
guillaume-chervet, line 5: Thank you @jafin for the Awesome PR.

When i run npm install, I have the error bellow.

image

packages/react/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@axa-fr/react-oidc",
   "version": "6.15.8",
   "private": false,
```
guillaume-chervet, line 4: Thank you @jafin, I have got more time today to check.
To publish to npm all files. You need to updated "files" 
`
  "files": [
    "dist",
    "bin",
    "src/oidc",
    "service_worker",  <--- add this
    "src/override",
    "README.md",
    "package.json",
    "package-lock.json"
  ],
`

guillaume-chervet, line 4: @jafin ,

The package.json in vanilla folder have to be adapted too.

I would like to set up back learna but I'am lacking of time : https://github.com/lerna/lerna
To set it up quickly, i copied file from react to vanilla in order to publish vanilla library.

packages/react/service_worker/utils/tokens.ts
```diff
@@ -0,0 +1,198 @@
...
+) {
+  const currentTimeUnixSecond = new Date().getTime() / 1000;
+  return Math.round(
+    expiresAt -
+      refreshTimeBeforeTokensExpirationInSecond -
+      currentTimeUnixSecond
+  );
+}
+
+function isTokensValid(tokens: Tokens | null) {
```
guillaume-chervet, line 33: Thank you so much for this refacto

packages/react/package.json
jafin: @guillaume-chervet thanks I'll investigate lerna and get back with an update.

guillaume-chervet: Vanilla lib just need and update package.json that copy and publish the new service worker folder already compiled from react folder.

jafin: Sorry, I misunderstood the intent of your comments.  I will amend then with the changes you have suggested.  Leaving lerna as a separate PR.

jafin: @guillaume-chervet I've amended the vanilla package to copy the compile oidcServiceWorker.js from React package.
It's a bit of a muddle since vanilla doesn't clarify that react/service_worker needs to be compiled. I think following up with a Lerna/workspaces to enable package dependency should reduce the complication. I'll continue to look as this as a separate PR improvement.

guillaume-chervet (approved):
I have test it, it'is perfect, may I merge it?
Thank you so much @jafin !

guillaume-chervet:
I merge it because my enterprise will rename AxaGuilDEv as axafrance this morning.
I will not know the result :p

Issue #974, opened 2023-03-01
Uncaught TypeError on page refresh. Failed to construct 'Request' with a RequestInit whose mode member is set as 'navigate'


jafin:
### Issue and Steps to Reproduce

There is no perceived functionality loss, just an exception being logged in the console.

### Versions
6.14.10

### Screenshots
![image](https://user-images.githubusercontent.com/127927/222071391-7fcbf9f0-b58e-4f87-894e-2a3cdc29ae5c.png)

#### Expected
No exception be raised in the console.

#### Actual
The exception is raised in the console, occasionally when a browser refresh is triggered via F5
Everything works, just a console exception is logged.

### Additional Details

I've been unable to repro this in the online demo.

Steps:
I am authenticated inside a react app.
When I F5 refresh the browser, the error will be emitted.
Now the error is for a request for the URL to reload the app.  So perhaps the handleFetch doesn't need to run for this.

```js
Uncaught (in promise) TypeError: Failed to construct 'Request': Cannot construct a Request with a RequestInit whose mode member is set as 'navigate'.
```

Suggestion:

```js  
//OidcServiceWorker.js
const handleFetch = async(event)=>{
  const originalRequest = event.request;
  if (originalRequest.mode === 'navigate')
     return;  //bail early
...
```

The TypeError: failed to fetch will still be logged, but the uncaught exception now doesn't occur.

- Installed packages:
N/A

rike422:
I am encountering the same problem.
It is occurring on version 6.14.12
As an additional note, it also occurs when multiple tabs are opened.

jafin:
@rike422  do you experience any ill effects from this? or just the error in the console?  Asking because if it's just the visible error, I'll propose a PR to implement the suggestion I mention above.

rike422:
@jafin 
We have not verified this in detail, but it seems to be generally functioning properly.
Currently, a lot of logs are left in error tracking tools, so it would be very helpful if measures are taken.

jafin:
I think this may actually relates to #982,  The property `service_worker_convert_all_requests_to_cors` set true should resolve this.

PR #985, opened 2023-03-07
fix(serviceworker): Copy prop service_worker_convert_all_requests_to_cors

### LINKED ISSUE #982: Authorization headers not being set when request mode is navigate

When trying to set the authorization headers in the service worker if the request mode is set as  'navigate' it produces an error and does not create the new request with the headers.

error: Uncaught (in promise) TypeError: Failed to construct 'Request': Cannot construct a Request with a RequestInit whose mode member is set as 'navigate'.

### Versions
6.15.2

### Screenshots

image

#### Expected

Set authorization headers in every request mode.

#### Actual

When request mode is 'navigate' it does not set authorization headers.

### Additional Details

Changing the original request mode before creating the new request worked for me, can this be implemented in any way in the service worker for future versions?

```
// OicdServiceWorker

handleFetch = async (event) => {
   ...
  originalRequest.mode = originalRequest.mode === "navigate" ? "cors" : originalRequest.mode;
  const newRequest = new Request(originalRequest, {
  ...
}
```

- Installed packages:

jafin:
Ensure the prop `oidcConfiguration.service_worker_convert_all_requests_to_cors`  is copied when constructing oidcConfiguration
May resolve #982 and #974 (when configured with the prop set true)

Question: What should the `default` value `service_worker_convert_all_requests_to_cors` ?
Setting it true would avoid the error being reported, would that be considered the default/preferred configuration?

### DIFF
--- a/packages/react/src/oidc/vanilla/initWorker.ts
+++ b/packages/react/src/oidc/vanilla/initWorker.ts
@@ -200,7 +200,14 @@ export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName
     const initAsync = async (oidcServerConfiguration, where, oidcConfiguration:OidcConfiguration) => {
         const result = await sendMessageAsync(registration)({
             type: 'init',
-            data: { oidcServerConfiguration, where, oidcConfiguration: { token_renew_mode: oidcConfiguration.token_renew_mode } },
+            data: {
+                oidcServerConfiguration,
+                where,
+                oidcConfiguration: {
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank very much you again @jafin. That a big bug ^^
I have some taks to finish this month and I think i will have more time to work on this library to clean and test it.

For this kid of library the default value should be the securest one. In this case 
 service_worker_convert_all_requests_to_cors=false.

guillaume-chervet (approved):

PR #1003, opened 2023-03-18
refactor(build): Don't run azure deploys or publish on PR's

jafin:
## Issue
When a PR is raised, the pipeline workflows attempt to run the Build/Publish and the Azure publish jobs.
These appear to fail often (always?).

## Change
Removed trigger on pull_requests for Azure and Publish workflows.
Leaving only trigger on push->master

## Additional note
Action `egorooj/skip-ci@main` may not be required anymore, see (https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs)

### DIFF
--- a/.github/workflows/azure-static-web-apps-icy-glacier-004ab4303.yml
+++ b/.github/workflows/azure-static-web-apps-icy-glacier-004ab4303.yml
@@ -4,14 +4,17 @@ on:
   push:
     branches:
       - master
-  pull_request:
-    types: [opened, synchronize, reopened, closed]
-    branches:
-      - master
 
 jobs:
+  skip_ci:
+    runs-on: ubuntu-latest
+    outputs:
+      canSkip: ${{ steps.check.outputs.canSkip }}
+    steps:
+      - id: check
+        uses: Legorooj/skip-ci@main
   build_and_deploy_job:
-    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.action != 'closed')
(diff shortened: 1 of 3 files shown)


.github/workflows/azure-static-web-apps-icy-glacier-004ab4303.yml
```diff
@@ -4,14 +4,17 @@ on:
   push:
     branches:
       - master
-  pull_request:
```
guillaume-chervet, line 7: hi @jafin , I would prefer to keep this, it allow me to test on many devices before merging.
I do not have solution for PR from fork. The best is as you set up bellow to skip it when PR comes from a fork :  !github.event.pull_request.head.repo.fork

.github/workflows/npm-publish.yml
```diff
@@ -6,8 +6,6 @@ name: React Oidc CI
 on:
   push:
     branches: [ master ]
-  pull_request:
```
guillaume-chervet, line 9: idem, I would like to keep it.

.github/workflows/azure-static-web-apps-icy-glacier-004ab4303.yml
jafin: @guillaume-chervet  All good, was just a suggestion. I was curious if there was an option on PR builds to get the builds to go green, so as a committer it was easy to ensure I didn't break stuff,  But if the way you have it setup now benefits in deployments then that makes sense to maintain it.  I'll close the PR

guillaume-chervet:
@jafin,
I will try to set up a flow that will be green for PR from Fork.

still thank you so much for your awesome help.

PR #1005, opened 2023-03-19
fix(modules): Fix issue with compilation of react package to ESmodules format

jafin:
## Bugfix
- Error issue introduced in e128be41573e158cc052e6ae1f08805adbcca058 when changed package.json to module but typescript still emit commonJS format.

This change sets typescript module type is ESModules so it emits to match package.json `"type":"module"`

chrome_UwdeRL5kX9

## Before this PR

React package will throw error `exports is not defined. in fetchtoken

## After this PR

Described error should be resolved.

### DIFF
--- a/packages/react/tsconfig.json
+++ b/packages/react/tsconfig.json
@@ -5,7 +5,7 @@
     "outDir": "dist",
     "baseUrl": "./",
     "jsx": "react-jsx",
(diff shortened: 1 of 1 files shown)


guillaume-chervet (approved):
thank you @jafin

Issue #1006, opened 2023-03-20
export 'OidcProvider' (imported as 'OidcProvider') was not found in '@axa-fr/react-oidc/dist/index' (module has no exports)


reeaazz:
### Issue and Steps to Reproduce

I tried to recreate this simple app but when i try to call few things from the module i get lots of errors (module has no exports) 

#### Actual
ERROR in ./src/App.js 20:30-42
export 'OidcProvider' (imported as 'OidcProvider') was not found in '@axa-fr/react-oidc/dist/index' (module has no exports)

ERROR in ./src/Layout/Header.jsx 13:6-17
export 'useOidcUser' (imported as 'useOidcUser') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/Layout/Header.jsx 62:10-21
export 'useOidcUser' (imported as 'useOidcUser') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/pages/Admin.jsx 9:30-40
export 'OidcSecure' (imported as 'OidcSecure') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/pages/Home.jsx 15:6-13
export 'useOidc' (imported as 'useOidc') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/pages/Home.jsx 19:6-20
export 'useOidcIdToken' (imported as 'useOidcIdToken') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/pages/Home.jsx 43:10-17
export 'useOidc' (imported as 'useOidc') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./src/pages/Home.jsx 43:19-33
export 'useOidcIdToken' (imported as 'useOidcIdToken') was not found in '@axa-fr/react-oidc' (module has no exports)

ERROR in ./node_modules/@axa-fr/react-oidc/dist/index.js 1:0-59
Module not found: Error: Can't resolve './FetchToken' in '/Users/amarrai/react-oidc-example-main/node_modules/@axa-fr/react-oidc/dist'
Did you mean 'FetchToken.js'?
BREAKING CHANGE: The request './FetchToken' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
[...]

PR #1009, opened 2023-03-21
fix(react): Fix issue where imports not found due to missing .js extension

jafin:
## Issue
When `module: ESNext` was introduced in 26c150be3222541bb8e3071c339843137ed9a845 it has broken imports for consumers.

Causing issues described in #1006 

## Resolution
If I understand correctly  https://github.com/microsoft/TypeScript/issues/42151  imports for esmodules should be written with a `.js` extension (there is also a recommendation to set this as the default in vscode etc) , as tsc (by design ?) will not alter import statements in .ts files even when tsconfig is `module: esnext`, 

This PR updates all the imports to have the .js extension.

## Alternatives

- https://github.com/Zoltu/typescript-transformer-append-js-extension

### DIFF
--- a/packages/react/src/App.tsx
+++ b/packages/react/src/App.tsx
@@ -1,12 +1,12 @@
 import React, { useReducer } from 'react';
 import { BrowserRouter, NavLink, Route, Routes } from 'react-router-dom';
 
-import { configurationIdentityServer } from './configurations';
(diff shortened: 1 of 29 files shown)


guillaume-chervet (approved):
again thank you so much @jafin

PR #1010, opened 2023-03-22
fix(imports): Add aliases for CRA app to resolve .js extensions 

jafin:
# Issue

When ESModule was added, one fix was to append .js to the imports to ensure when code is transpiled it is correctly referenced.
This now cause an error for the react demo app in package/react. 

Relates to #1006 

# Proposed Fix

Adjust CRA to to set `resolve.extensionAlias` to fallback to check for a matching `.ts` when looking for `.js` files (I'm assuming when running CRA/webpack the transpile doesn't write the files to disk, therefore the import ../.js file errors. 

Because CRA doesn't allow access to webpack config, `craco` is used.

The other minor change was an import issue  in Profile, adjusted  the import/export to resolve.


guillaume-chervet:
I waswrong it not link with the boostrap update

guillaume-chervet:
It work now on my personal computer :/

PR #1011, opened 2023-03-22
fix(ci): skip task with secrets from for PullRequest

guillaume-chervet:
hi @jafin , this is not perfect but it should allow to run unit test from fork pullrequest.

We may set up a working branch in order to autodeploy demo and alpha-beta after a merge.


jafin:
@guillaume-chervet thanks a good improvement!

PR #1012, opened 2023-03-22
fix(oidc): wait for valid tokens before capturing value of accessToken

conleos-lrohr:
## Before this PR

Scenario: you have logged into an app and close it to open it again later/the next day. The time you open it again, the idToken/accessToken have been expired, but the refreshToken is still valid. New tokens are generated with the refresh token, but when trying to get the userInfo at the same time, the old accessToken is used.

This happens, because the (possibly invalid) value of `oidc.tokens.accessToken` has been captured in the variable `accessToken`, afterwards it is checked whether the tokens are valid, but in the end the captured value of access token is used to make the userInfo request.

## After this PR

The newly retrieved accessToken is used for the userInfo request.

### DIFF
--- a/packages/react/src/oidc/vanilla/user.ts
+++ b/packages/react/src/oidc/vanilla/user.ts
@@ -5,17 +5,18 @@ export const userInfoAsync = async (oidc) => {
     if (oidc.userInfo != null) {
         return oidc.userInfo;
     }
+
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you very much @conleos-lrohr .

guillaume-chervet (approved):

PR #1018, opened 2023-03-27
fix(oidc): null error when no valid token is somehow present

### LINKED ISSUE #1017: fetchWithToken crashes with a null error when no tokens are present

Not 100% certain, but the issue started occuring after a failed attempt at refreshing the token, I wanted to reproduce an authentication error page because I'm building custom components for them.
The way I did is by temporarily taking out the refresh_token grant type from the backend whilst it's attempting to refresh the token, it fails, restart the backend with the grant type enabled, and now it's stuck in the error loop (and it won't recover either by resetting cache)

The specific offending line is:
```js
  const accessToken = getValidToken.tokens.accessToken;
```

Considering that this method already doesn't mandate a token it's presence in the first place, could there be a null-coalescing operator placed here?

### Versions
6.16.7

### Screenshots
2023-03-27 19_30_18-DevTools - localhost_3000_app_dashboard

#### Expected
The application should not crash

#### Actual
The application crashes when attempting to send a fetch request with a token

### Additional Details

- Installed packages:

Thodor12:
fixes #1017

### DIFF
--- a/packages/react/src/oidc/FetchToken.tsx
+++ b/packages/react/src/oidc/FetchToken.tsx
@@ -22,7 +22,7 @@ const fetchWithToken = (fetch: Fetch, getOidcWithConfigurationName: () => Vanill
 
   // @ts-ignore
   const getValidToken = await oidc.getValidTokenAsync();
-  const accessToken = getValidToken.tokens.accessToken;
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you @Thodor12

Issue #1016, opened 2023-03-27
Release v6.16.8 doesn't copy service worker in public directory anymore


AnimaLupi:
### Issue and Steps to Reproduce

`npm install  @axa-fr/react-oidc@6.16.8`

The file `OdicServiceWorker.js` isn't copied to the `public` directory (but previous version is removed).

### Versions
6.16.8

#### Expected
The service worker is installed, updated or preserved

#### Actual
The service worker is removed

The problem isn't present in version 6.16.7.

AnimaLupi:
Thank you for the fast feedback!

Yes, I'm using create react app with `react-scripts` at version 4.0.3 and some (minimal) customization with `craco` (`craco-plugin-env`, `craco-less`).

If you think that could be the cause of the issue (or some other dev dependency used in the build chain) I can provide the full package.json and/or the craco configuration (which is limited to those two plugins, though).

Zamaletdinov:
I experience the same issue at the moment with vite - migrated to it from CRA a long while ago

NickCain:
I also have the same problem since upgrading, also it seems to not be fixed by rolling back for me either.

Zamaletdinov:
@NickCain I have downgraded back to 6.15.5 and the file has been regenerated

NickCain:
@Zamaletdinov Thanks for your reply. Unfortuately even that version doesnt generate it for me now, i've even removed all node_modules and yarn cache

AnimaLupi:
Thank you for the fix, now updating the package preserve/updates the `OidcServiceWorker.js` file properly.

guillaume-chervet:
Hi @AnimaLupi , may we close the issue?

AnimaLupi:
> Hi @AnimaLupi , may we close the issue?

For what I'm experiencing now, it's fixed and can be closed, thank you! :+1:

guillaume-chervet:
Thank you very much for your issue @AnimaLupi .

PR #1020, opened 2023-03-29
Resolve issue #1016 by updating Vite service worker config

dcopestake:
Resolve issue #1016 by updating Vite service worker config to output CommonJS instead of ES.

## Before this PR

Since the migration of the `OidcServiceWorker.js` code to TypeScript, there is no `OidcServiceWorker.js` file being generated, instead we have an ES module called `OidcServiceWorker.mjs` instead. This means the `copy.cjs` script silently fails to copy the script over to the `public` folder because of this code:

```
const serviceworkerFilePath = path.join(__dirname, "..", "dist/service_worker", serviceworkerFilename);
if (fs.existsSync(serviceworkerFilePath)) {
...
```

## After this PR

We get a `OidcServiceWorker.js` in the `dist` folder and the `copy.cjs` script should work as before.


guillaume-chervet (approved):
thank you very much @dcopestake

Issue #1029, opened 2023-04-11
OidcKeepAliveServiceWorker.json requests failing following pull request #967


dcopestake:
### Issue and Steps to Reproduce

Following the merge of a previous pull request #967 from me to solve issue #862, unfortunately this introduced an error where the `OidcKeepAliveServiceWorker.json` requests were failing because of the `GET` fetch requests not being allowed to contain a body. This results in the SW throwing the exception `TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.`.

Should be showing for everyone I think so reproducing should be relatively straightforward! Apologies.

### Versions

v6.16.12

wooandoo:
the problem is still present on version "@axa-fr/react-oidc": "6.16.14", but the error is now logged as:

```
OidcKeepAliveServiceWorker.json?minSleepSeconds=150 net::ERR_FAILED

TypeError: Failed to fetch
    at keepAlive (initWorker.ts:114:25)
```

anonkey:
> the problem is still present on version "@axa-fr/react-oidc": "6.16.14", but the error is now logged as:
> 
> ```
> OidcKeepAliveServiceWorker.json?minSleepSeconds=150 net::ERR_FAILED
> 
> TypeError: Failed to fetch
>     at keepAlive (initWorker.ts:114:25)
> ```

Creating an OidcKeepAliveServiceWorker.json with "{}" into it fixed this problem for me

hsul-git:
Same problem here:

`initWorker.ts:114     GET http://127.0.0.1:3000/OidcKeepAliveServiceWorker.json?minSleepSeconds=240 net::ERR_FAILED`

PR #1030, opened 2023-04-11
Fix issue #1029 by moving minimum sleep duration to query string

dcopestake:

As mentioned in issue #1029, `OidcKeepAliveServiceWorker.json` requests were failing because of the `GET` fetch requests not being allowed to contain a body. 

## Before this PR

SW throws exception `TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.` for keep alive requests.

## After this PR

Keep alive requests succeed properly.


guillaume-chervet:
Thank you again @dcopestake !

Issue #1038, opened 2023-04-24
Access tokens in service worker mode


krzempekk:
### Issue and Steps to Reproduce

When using service worker, tokens are not accessible to client code. This is great from security perspective, but we have some use cases in which we need to have access to tokens and still want to utilise service workers (to take advantage of tokens synchronisation between tabs, etc.):
- micro frontends rendered inside and iframe, which rely on parent app for authentication and use its tokens for requests
- internal tools that need to display access token in dev environments.

Configuration option to turn off tokens hiding would be great for us and tokens would still be hidden by service worker by default.

### Versions

6.16.14

#### Expected

There is boolean configuration option (in `OidcConfiguration`, could be named `service_worker_hide_tokens`, `true` by default) to turn off hiding tokens by service worker.

#### Actual

Tokens are always hidden when using service workers.

guillaume-chervet:
Hi @krzempekk, thank you for your issue.

you can disable service worker mode in development just by not configuring the serviceworker file in development you will have the behavior you need.

krzempekk:
Hi @guillaume-chervet!
Thank you for response. Yes, I understand that service worker can be not configured for this behaviour, but in our case we need to use service worker (to have token synchronisation, synchronised renewal, etc.) and still have access to tokens.
As I have said, we have production use cases for that, so it's not only development.

guillaume-chervet:
Hi @krzempekk , I am not sure but does localStorage mode do the job you need? It is the same behavior than Service worker (except you need to inject accessToken manually in each fetch) and you have access token not offuscated.

krzempekk:
Hi @guillaume-chervet! 
Unfortunately localStorage is not an option for us - we have requirement to store tokens only in-memory, as service worker does.

guillaume-chervet:
Thank you @krzempekk , i think i understand your need.
This is possible to add, but the configuration will able to be set from trusteddomains.js for security readon.

krzempekk:
That would be perfectly fine for us to set this configuration from `OidcTrustedDomains.js`. Thank you @guillaume-chervet, looking forward to this :)

darkowic:
@guillaume-chervet just to be clear: in case we decide to use this library (which will probably happen), we will be willing to help resolve the issues we are raising :) I hope you are open to collaboration and enhancements propositions like this one :)

guillaume-chervet:
Hi @darkowic thank you for your message.

How much time do you have?
[...]

PR #1046, opened 2023-05-03
feat(oidc): allow service worker to not hide acces_token

guillaume-chervet:
Feature for #1038.

Need to add unit test.


packages/react/public/types.d.ts
```diff
@@ -47,6 +50,7 @@ export type OidcConfig = {
     oidcConfiguration?: OidcConfiguration;
     sessionState?: string | null;
     items?: MessageData;
+    hideAccessToken: boolean;
```
darkowic, line 53: `hideAccessToken` is `true` by default. Should we call this the opposite: `showAccessToken`?

packages/react/service_worker/OidcServiceWorker.ts
```diff
@@ -274,11 +279,14 @@ const handleMessage = (event: ExtendableMessageEvent) => {
       nonce: null,
       status: null,
       configurationName,
+      hideAccessToken: trustedDomainsHideAccessToken[configurationName],
```
darkowic, line 282: @krzempekk we need to evaluate this.

@guillaume-chervet we need to check if the trusted domains are something we can dynamically generate based on our deployment target. In our deployment model, we follow the: build once and deploy to many environments (dev/staging/prod). Therefore, anything hardcoded at the build time is a problem.

guillaume-chervet: Thank you for your feedback @darkowic.

I'am waiting for your response :)

HSULGIT: How can we set the access token to be shown?

PR #1051, opened 2023-05-10
fix(react-oidc): add useCallback for fetch function to memoized it

### LINKED ISSUE #1023: `useOidcFetch` as a `useEffect` dependency

I want to use the `fetch` returned by the `useOidcFetch` hook in a [`useEffect`](https://react.dev/reference/react/useEffect) setup function. My linter warns me that I ought to list `fetch` as one of the dependencies of the effect: 

> React Hook `useEffect` has a missing dependency: `'fetch'`. Either include it or remove the dependency array  `react-hooks/exhaustive-deps`

However, it seems the value returned by `useOidcFetch` differs from call to call, so listing `fetch` in the dependencies array of  `useEffect` causes the effect to run in an infinite loop - meaning the same call to `fetch` is made over and over, wasting network resources.

For now, I am omitting `fetch` from the dependency array, and ignoring the linter error, but it would be nice if this would work out of the box.

```ts
import React, { useEffect, useState } from 'react';
import { useOidcFetch } from '@axa-fr/react-oidc';

// Mount this component inside <OidcProvider> and <OidcSecure>
export const Example = () => {
  const { fetch } = useOidcFetch();
  const [reponse, setResponse] = useState(null as any);
  useEffect(() => {
    fetch('https://api.example.com') // set this to a URL configured to allow CORS requests
      .then(setResponse)
      .then(() => console.log(`Fetch finished at ${Date()}`))
      .catch(console.error);
  }, [fetch, setResponse]); // remove fetch to prevent repeated requests, but then ESLint warns about react-hooks/exhaustive-deps
  return <div>See console log</div>;
};
```

### Versions

`@axa-fr/react-oidc` 6.16.10

#### Expected

The effect should only be run once.

#### Actual

The effect is run repeatedly, so long as `fetch` is in the dependency array.

guillaume-chervet:
in order to fix #1023

### DIFF
--- a/packages/react/src/oidc/FetchToken.tsx
+++ b/packages/react/src/oidc/FetchToken.tsx
@@ -1,3 +1,5 @@
+import { useCallback } from 'react';
+
 import { VanillaOidc } from './vanilla/vanillaOidc.js';
 
 export type Fetch = typeof window.fetch;
@@ -47,7 +49,14 @@ export const withOidcFetch = (fetch: Fetch = null, configurationName = defaultCo
 export const useOidcFetch = (fetch: Fetch = null, configurationName = defaultConfigurationName) => {
   const previousFetch = fetch || window.fetch;
   const getOidc = VanillaOidc.get;
-  const getOidcWithConfigurationName = () => getOidc(configurationName);
-  const newFetch = fetchWithToken(previousFetch, getOidcWithConfigurationName);
-  return { fetch: newFetch };
+
+  const memoizedFetchCallback = useCallback(
+      (arg1, arg2) => {
+        const getOidcWithConfigurationName = () => getOidc(configurationName);
+        const newFetch = fetchWithToken(previousFetch, getOidcWithConfigurationName);
+        return newFetch(arg1, arg2);
+      },
+      [previousFetch, configurationName],
+  );
+  return { fetch: memoizedFetchCallback };
 };


packages/react/src/oidc/FetchToken.tsx
```diff
@@ -47,7 +49,14 @@ export const withOidcFetch = (fetch: Fetch = null, configurationName = defaultCo
...
   const getOidc = VanillaOidc.get;
-  const getOidcWithConfigurationName = () => getOidc(configurationName);
-  const newFetch = fetchWithToken(previousFetch, getOidcWithConfigurationName);
-  return { fetch: newFetch };
+
+  const memoizedFetchCallback = useCallback(
+      (arg1, arg2) => {
+        const getOidcWithConfigurationName = () => getOidc(configurationName);
+        const newFetch = fetchWithToken(previousFetch, getOidcWithConfigurationName);
+        return newFetch(arg1, arg2);
```
jasonaowen, line 57: I'm afraid this is a breaking change, as it makes the previously-optional second parameter required. Upgrading to the version with this change breaks existing code:

```
TS2554: Expected 2 arguments, but got 1.
  > 32 |     fetch(url)
```

You can also see this in the compiled type definition `node_modules/@axa-fr/react-oidc/dist/FetchToken.d.ts`:

```ts
export declare const useOidcFetch: (fetch?: Fetch, configurationName?: string) => {
    fetch: (arg1: any, arg2: any) => Promise<Response>;
};
```

I think you'll need to do some type trickery to handle this correctly - maybe something like `(resource: any, options?: any) => ...`, although I regret the `any`s - but I don't have that fully tested and wanted to flag this as something you might want to consider reverting in the meantime.

jasonaowen (commented):
Thank you for working on this, @guillaume-chervet! I apologize I didn't get a chance to review it sooner.

PR #1080, opened 2023-07-03
fix(vanilla-demo): proper fetch injection

kosciolek:
`getFetch` is `private getFetch: () => Fetch;`, which means it should be a function that returns `fetch`, not `fetch` itself.

If you try to run `vanilla-demo` using the local `@axa-fr/vanilla-oidc`, it will throw with "fetch is not a function":

```sh
cd packages/react
npm run i
cd ..
cd vanilla
npm run i
npm link
cd ..
cd vanilla-demo
npm run i
npm link @axa-fr/vanilla-oidc
npm run start
```

### DIFF
--- a/packages/vanilla-demo/src/index.tsx
+++ b/packages/vanilla-demo/src/index.tsx
@@ -88,7 +88,7 @@ export const configuration = {
 
 const href = window.location.href;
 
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you @kosciolek fior your help :)

PR #1081, opened 2023-07-03
WIP: delegate SW registration to the user

kosciolek:
just an exploratory PR, feel free to close it at will

### DIFF
--- a/packages/react/src/oidc/vanilla/initWorker.ts
+++ b/packages/react/src/oidc/vanilla/initWorker.ts
@@ -151,7 +151,7 @@ const sendMessageAsync = (registration) => (data) => {
     });
 };
 
-export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName) => {
+export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName, noAutoRegister) => {
     if (typeof window === 'undefined' || typeof navigator === 'undefined' || !navigator.serviceWorker || !serviceWorkerRelativeUrl) {
         return null;
     }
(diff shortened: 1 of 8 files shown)


packages/react/src/oidc/vanilla/initWorker.ts
```diff
@@ -188,21 +187,6 @@ export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName
         return await registration.unregister();
     };
 
-    registration.addEventListener('updatefound', () => {
```
kosciolek, line 191: this code reloads the page if an update to a service worker is detected, is it really needed? if the user manages the service worker on his own, it probably should be on his side

guillaume-chervet: Yes, I know,

may be another property to disable this default behavior would be a nice idea.
The difficulty, is link to the time to leave request which maintain the old service worker to life and block updates.

I does not found a cleaner default behavior way. Many people does not understand well service worker so I prefer a default page realoard in production than all production page broken. 

May be a clean defaut and safer way can be found.

guillaume-chervet:
Please continue You PullRequest.
I like it ! and thank you very much for your help :)

guillaume-chervet:
hi @kosciolek ,

I'am very sorry @kosciolek  for the delay. I was too busy :'(
I very like your PullRequest. Do you have time to update it?

darkowic:
Hi @guillaume-chervet! This PR was discontinued as we deprioritized this functionality on our side. Feel free to close this and we will eventually continue if we find this feature needed.

esomore:
I would like to have this in My PWA using Vite. I have to use it without the service worker functionality since it is breaking my Periodic SW Updates (https://github.com/vite-pwa/vite-plugin-pwa/blob/main/docs/frameworks/react.md#periodic-sw-updates)

PR #1106, opened 2023-07-24
refactor(post-install): Alter post-install script to not take a dependency on cpy

jafin:
The `post-install` script took a dependency on lib `cpy`   Because this is not marked as a dependency it will error for a consumer.
This change removes the external dependency on cpy for the post-install script. 

Fixes issue found in 6.25.4-alpha952 where if consumer doesn't have `cpy` as an existing dependency, it will fail to install the package.

```
../node_modules/@axa-fr/vanilla-oidc postinstall$ node ./bin/post-install.mjs
│ node:internal/errors:490
│     ErrorCaptureStackTrace(err);
│     ^
│ Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'cpy' imported from \node_modules\.pnpm\@axa-fr+vanilla-oidc@6.25.4-a…  
│     at new NodeError (node:internal/errors:399:5)
│     at packageResolve (node:internal/modules/esm/resolve:889:9)
```


guillaume-chervet (approved):
Thank you again @jafin .
I think it is the simplest solution.

PR #1107, opened 2023-07-24
ci(version): Add a period between version alpha|beta and run_number

jafin:
Introduce a period between the `alpha|version` meta tag and the `build number`.
Although not strictly semver (they appear to use a +) apparently NPM will strip the + metadata though so that is not ideal.
It appears common for libs to use the format  `x.xxx-beta.xx`

This change aligns to that format.

## Before this PR

Version would be `1.4.0-alpha5`

## After this PR

Version will be `1.4.0-alpha.5`


guillaume-chervet (approved):
Thank you again @jafin

PR #1149, opened 2023-10-04
Reclaming all clients from SW after hard reload

### LINKED ISSUE #1098: Hard-reload in browser unregister ServiceWorker

When hard-reload in browser, the ServiceWorker seems to be unregistered. He doesn't intercept the calls to inject the token meaning that all the calls will failed with `Unauthorized` error.
After a simple refresh in browser, everything works fine.
This can be reproduce in your [demo react.](https://black-rock-0dc6b0d03.1.azurestaticapps.net/)

### Versions
v6.24.1
### Screenshots
image
image
#### Expected
ServiceWorker to intercept and inject token after a hard-reload in browser
image

#### Actual
ServiceWorker isn't injecting token after a hard-reload

Thanks!:)

VladimirPlatonenko:
Detecting situation when service worker installed but does not control application(this is valid situation after hard refresh) and reclaiming control in response. Since this has to be done from within service worker itself have to send separate message to it.

## Before this PR
After Hard refresh we lose Bearer token injection for request. As a result you get 401 for most of request.

## After this PR
All works smooth

should fix #1098

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -189,6 +189,8 @@ export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName
 
     try {
         await navigator.serviceWorker.ready;
+        if (!navigator.serviceWorker.controller)
+            await sendMessageAsync(registration)({ type: 'claim' });
     } catch (err) {
         return null;
     }
--- a/packages/oidc-client-service-worker/src/OidcServiceWorker.ts
+++ b/packages/oidc-client-service-worker/src/OidcServiceWorker.ts
@@ -283,6 +283,10 @@ const handleFetch = async (event: FetchEvent) => {
 const handleMessage = (event: ExtendableMessageEvent) => {
   const port = event.ports[0];
   const data = event.data as MessageEventData;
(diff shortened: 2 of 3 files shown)


packages/oidc-client/src/initWorker.ts
```diff
@@ -189,6 +189,8 @@ export const initWorkerAsync = async(serviceWorkerRelativeUrl, configurationName
 
     try {
         await navigator.serviceWorker.ready;
+        if (!navigator.serviceWorker.controller)
```
guillaume-chervet, line 192: Hi @VladimirPlatonenko ,

Thank you so much for your help!

Very sorry for the delay. I will try to test it as soon as i can.
The modification have to be done in the service worker packages.

Here, in this file, it won't affect the published library.

VladimirPlatonenko: I moved changes to ts file for service

guillaume-chervet:
Very Sorry for the delay @VladimirPlatonenko.
I have tested it. I do not understand how it works, but it works !

Thank you very much @VladimirPlatonenko :)

guillaume-chervet (approved):

PR #1196, opened 2023-11-16
fix: add URL normalization for getCurrentDatabaseDomain endpoints comparison

### LINKED ISSUE #1189: Normalize URLs in OidcServiceWorker.js

Currently URLs in comparisons are not normalized, which can cause certains calls to fail; e.g. 

```ts
url === oidcServerConfiguration.tokenEndpoint
```

this will return false if one URL contains a port and the other does not (some IDPs include a port).

I can create a merge request if needed; we could use a package `normalize-url` to solve it, or implement only a subset of the rules.

### Versions

`@axa-fr/oidc-client-service-worker@7.7.3`

meesvandongen:
fixes #1189

### DIFF
--- a/packages/oidc-client-service-worker/src/utils/__tests__/domains.spec.ts
+++ b/packages/oidc-client-service-worker/src/utils/__tests__/domains.spec.ts
@@ -1,94 +1,155 @@
 import { describe, expect, it } from 'vitest';
 
 import { openidWellknownUrlEndWith } from '../../constants';
-import { checkDomain, getCurrentDatabaseDomain } from '..';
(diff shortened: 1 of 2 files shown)


meesvandongen:
I took `normalize-url` and removed some unneeded rules, I can probably bring it down even further by ignoring things that don't make sense for a user to input; for example the trim

guillaume-chervet:
Thank you @meesvandongen , i test it right now.

guillaume-chervet (approved):

guillaume-chervet:
Thank you very much @meesvandongen

PR #1198, opened 2023-11-23
fix: make sure URL is normalized when comparing with token endpoint a…

### LINKED ISSUE #1189: Normalize URLs in OidcServiceWorker.js

meesvandongen:
…nd revocation endpoint

fixes #1189


guillaume-chervet (approved):
Than you @meesvandongen it is much cleaner, I like it !

I have tested it and it works!

PR #1200, opened 2023-11-29
Update README.md

felixroos:
fix service worker command to work with oidc-client standalone


guillaume-chervet (approved):
Thank you @felixroos !

PR #1237, opened 2023-12-09
fix: make locks work only ifAvailable

abhilashlr:

### DIFF
--- a/packages/oidc-client/src/renewTokens.ts
+++ b/packages/oidc-client/src/renewTokens.ts
@@ -34,7 +34,8 @@ export async function renewTokensAndStartTimerAsync(oidc, refreshToken, forceRef
     if(configuration.storage === window.sessionStorage && !serviceWorker) {
         tokens = await syncTokens(oidc, refreshToken, forceRefresh, extras);
     } else {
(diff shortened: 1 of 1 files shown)


packages/oidc-client/src/renewTokens.ts
```diff
@@ -34,7 +34,8 @@ export async function renewTokensAndStartTimerAsync(oidc, refreshToken, forceRef
     if(configuration.storage === window.sessionStorage && !serviceWorker) {
         tokens = await syncTokens(oidc, refreshToken, forceRefresh, extras);
     } else {
-        tokens = await navigator.locks.request(lockResourcesName, async (lock) => {
+        tokens = await navigator.locks.request(lockResourcesName, { ifAvailable: true }, async (lock) => {
```
guillaume-chervet, line 37: I think it will require some little ajustement. But let published like it as an alpha. Thank you @abhilashlr !

guillaume-chervet (approved):

packages/oidc-client/src/renewTokens.ts
abhilashlr: Sure @guillaume-chervet ! Let me know what needs to be changed, I can work on it :)

PR #1311, opened 2024-02-28
fix(oidc): when having multiple iframes, only the last one gets resolved

### LINKED ISSUE #1309: multiple parallel silent signin: iframes are not removed before timeout
Let's say that you need to get 4 access tokens silently, only the last iframe gets removed when its access_token is acquired, the other 3 are removed at timeout (defined by the value of `configuration.silent_login_timeout`) even though their access_token are also acquired before the timeout occured.

## Steps to reproduce
```javascript
// assume I have conf1, conf2, conf3 and conf4 that represent the OidcConfiguration of my 4 OidcClient
// silent_login_timeout in each conf = 60000
const client1 = OidcClient.getOrCreate(() => window.fetch)(conf1, 'client1');
const client2 = OidcClient.getOrCreate(() => window.fetch)(conf2, 'client2');
const client3 = OidcClient.getOrCreate(() => window.fetch)(conf3, 'client3');
const client4 = OidcClient.getOrCreate(() => window.fetch)(conf4, 'client4');

// this will perform 4 silent signin at the same time
await Promise.all([
    client1,
    client2,
    client3,
    client4
].map((client) => client.loginAsync(undefined, undefined, false, undefined, true))
);
```

### Versions
7.19.0

#### Expected
- all clients' `loginAsync()` are resolved as soon as they acquire their access_token
- all clients' iframe are removed as soon as they acquire their access_token

#### Actual
- client1, client2 and client3 are rejected due to timeout (but they got their access_token and the serviceworker saved them)
- client1, client2 and client3 associated iframes are removed only after 60sec.
- client4 works as expected

### Additional Details
I find out that [this line](https://github.com/AxaFrance/oidc-client/blob/db13d8bdfbd49e564452461aeefc6e3024ce4435/packages/oidc-client/src/silentLogin.ts#L60C25-L60C58) returns `false` for client1, client2 and client3 so the `data` posted by the iframes cannot be parsed and then `_silentLoginAsync` cannot be resoled.

For client4 (the last one in `Promise.all()`), that line returns `true`, the data posted by its iframe get parsed and `_silentLoginAsync` is resolved as expected.

astrinxit66:
fix #1309

### DIFF
--- a/packages/oidc-client/src/silentLogin.ts
+++ b/packages/oidc-client/src/silentLogin.ts
@@ -44,7 +44,7 @@ export const _silentLoginAsync = (configurationName:string, configuration:OidcCo
         }
         const link = configuration.silent_login_uri + queries;
         const idx = link.indexOf('/', link.indexOf('//') + 2);
-        const iFrameOrigin = link.substr(0, idx);
+        const iFrameOrigin = link.substring(0, idx);
         const iframe = document.createElement('iframe');
         iframe.width = '0px';
         iframe.height = '0px';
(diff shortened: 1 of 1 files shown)


guillaume-chervet:
Thank you very much @astrinxit66 
I will merge it, it will be published as an alpha.

PR #1320, opened 2024-03-13
fix: normalize userinfo endpoint in service worker

meesvandongen:
## Before this PR

User info with port will not replace the access token.

## After this PR

User info with port will now properly forward the access token.


guillaume-chervet (approved):
Thank you so much @meesvandongen !

PR #1401, opened 2024-07-10
fix (demo): fix 'container' classname.

jafin:
`classname` may have accidently been modified to `contaicner` fixed with `container`.


guillaume-chervet (commented):
Thank you again @jafin

PR #1406, opened 2024-07-10
build(pipeline): Update actions to non-deprecated versions. bump node version to 20.  Run tests in non-wait mode

jafin:
- Update github actions to later versions (avoids the deprecated warning with node 18)
- Set Node version to `20` (from `18`) in pipeline
- modify test action in `package.json` to run in non-watch mode, so tests can be triggered from root workspace with `pnpm test` . 
  - simplify pipeline to just be `pnpm test`


guillaume-chervet (approved):
Thank you so much @jafin

PR #1411, opened 2024-07-13
build(job): Update job scripts to v4. (earlier versions are marked deprecated)

jafin:
- Updates job agents to v4, earlier versions are marked deprecated

> The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/checkout@v3, actions/setup-node@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

Note: This does not cover `CI.yml`, as that has been done in lint PR.


guillaume-chervet (approved):
Thank you again @jafin

PR #1413, opened 2024-07-13
tests(domain): Be more explicit about intent of test

jafin:
Tests are verifying no exception is thrown, not that the response is void. Changed to be more explicit about intent.


guillaume-chervet (approved):
Thank you @jafin

PR #1414, opened 2024-07-13
fix (FetchToken): Fix FetchToken param to be boolean instead of false (release)

jafin:
Change type of param to boolean (default: false) to resolve compile error:

```text
packages/react-oidc prepare: src/FetchToken.tsx:12:48 - error TS2345: Argument of type 'boolean' is not assignable to parameter of type 'false'.
packages/react-oidc prepare: 12   const newFetch = oidc.fetchWithTokens(fetch, demonstrating_proof_of_possession);
```

https://github.com/AxaFrance/oidc-client/actions/runs/9913967120/job/27392088934#step:6:121


guillaume-chervet (approved):
Thank you @jafin again !

PR #1416, opened 2024-07-15
chore(codestyle): apply prettier to packages

wermanoid:
## A picture tells a thousand words
There is a `prettier` and `.prettierrc.cjs` in repo

## Before this PR
It was not applied to code, so any change leads to massive file changes

## After this PR
Just applied code formatting for packages, so less annoying and unexpected file changes

## How to reproduce
run `pnpm prettier --write --cache packages`

PR #1417, opened 2024-07-15
fix(oidc): clenup client internal promises when requests are finished…

wermanoid:
… (release)

## Before this PR
In case if some request fails `class Oidc` used to keep promises internally with rejected state, so that leads to situation, that rejected promise is always returned for any sequential method call. That prevents to recover app connection to IDP after connection errors. (Mostly happens when internet disappears for some time or server was recently shut down for any reason) 

## After this PR
Oidc could execute method calls correctly, when environment is up and running after failure


wermanoid:
Might be good to review after https://github.com/AxaFrance/oidc-client/pull/1416

wermanoid:
better to review after https://github.com/AxaFrance/oidc-client/pull/1416

PR #1419, opened 2024-07-17
fix(eslint): Fix ESLint to run prettier

jafin:
Fix Eslint to run prettier.

The existing eslint config is incorrect and was not running the prettier validation rules.

This fails the lint step now and is expected, since all the prettier violations are detected.  PR #1416 should resolve them.


guillaume-chervet (approved):
Thank you @jafin you are helping so much!

guillaume-chervet (approved):

PR #1421, opened 2024-07-21
chore(lint): Fix lint warnings. (release)

jafin:
- Exclude `prettierrc.cjs` from lint
- lint prettier format autofix and adjust signature of Function
- null coalescing chain.


guillaume-chervet (approved):
Thank you very much again @jafin

PR #1423, opened 2024-07-22
fix(oidc): record tabIds when service worker receives init event (release)

krzempekk:
Fix from https://github.com/AxaFrance/oidc-client/pull/1402 has one issue - if `allowMultiTabLogin` flag is true and second tab is opened (when first tab is already authenticated), trying to initiate token renewal (using refresh token) from second tab fails. This is because previously tabId was saved to database only when login flow was initiated from given tab. This fix adds registering of tab id each time service worker receives `init` event.


krzempekk:
@guillaume-chervet could you take a look? This is follow-up for https://github.com/AxaFrance/oidc-client/pull/1402

guillaume-chervet (approved):
Thank you @krzempekk

PR #1427, opened 2024-07-26
fix(oidc): Add null as union type into arguments for login, logout, renewTokens

ShimpeiKamiguchi:

The TS2322 error occurs in `ReactOidc.tsx` because the type of the extras parameter only accepts `StringMap`. 
I think it should include `null` as a union.

## Before this PR

The TS2322 error occurs in `ReactOidc.tsx`

## After this PR

will prevent the error from occurring


guillaume-chervet (approved):
Thank you very much @ShimpeiKamiguchi

PR #1429, opened 2024-07-29
fix(oidc): Add undefined as union type into arguments for login, logout, renewTokens

ShimpeiKamiguchi:
# A picture tells a thousand words

I've modified default value as null for `extras` parameters in login, logout, renewTokens  https://github.com/AxaFrance/oidc-client/pull/1427
But I realized that they shoud be `undefined` because loginAsync,logoutAsync,renewTokensAsync accept `extras` as option.

# Before this PR

TS2345 erro occur in `ReactOidc.tsx`

# After this PR

will prevent the error from occurring


guillaume-chervet (approved):
Thank you @ShimpeiKamiguchi 
I will published it as an alpha

PR #1435, opened 2024-08-08
fix(logout): should work correctly with multiple tabs (release)

wermanoid:
## Before this PR
If you have `allowMultiTabLogin` and you have few tabs opened, you can face issue with logout function.

As I found, after logout there are still couple of messages sent from client to serwiceWorker (init and setState, at least).
This leads to case, when `currentDatabase.state` is changed after you passed logout function, but before you get redirected back to login screen. (At lease I think this is a reason. If you have better ideas about how to fix this - please comment, this is interesting).

So when you try to login after logout, you receive 400 error from OAuth2 server, because client sends request with some data from previous session (which was closed obviously).
 
## After this PR
You can loggin back to app from any tab without errors

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -57,6 +57,18 @@ export const defaultServiceWorkerUpdateRequireCallback =
     location.reload();
   };
 
+const getTabId = (configurationName: string) => {
+  const tabId = sessionStorage.getItem(`oidc.tabId.${configurationName}`);
+
+  if (tabId) {
+    return tabId;
+  }
+
+  const newTabId = globalThis.crypto.randomUUID();
+  sessionStorage.setItem(`oidc.tabId.${configurationName}`, newTabId);
+  return newTabId;
+};
+
 const sendMessageAsync =
   registration =>
   (data): Promise<any> => {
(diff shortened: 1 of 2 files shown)


packages/oidc-client/src/initWorker.ts
```diff
@@ -72,23 +84,13 @@ const sendMessageAsync =
         messageChannel.port1.close();
         messageChannel.port2.close();
       };
-      registration.active.postMessage(data, [messageChannel.port2]);
+      registration.active.postMessage({ ...data, tabId: getTabId(data.configurationName) }, [
```
wermanoid, line 87: this was very annoying that some actions has `tabId` and some are missing. I decided to unify this piece, so any interaction between client(s)/serviceWorker will hold ID for identification which tab performs specific action

guillaume-chervet: Thank you @wermanoid , very sorry for the delay. I am in holidays :)

guillaume-chervet: I can publish it as an alpha if you want ?

wermanoid: sure, let's do it :)

thanks!

guillaume-chervet: Published as an alpha @wermanoid thank you !

Issue #1439, opened 2024-08-23
Configuration option `service_worker_register` do not work as documented


thiscode:
### Issue and Steps to Reproduce

According to the instructions, you should use the option `service_worker_register` to inject your own logic loading the service worker. However, this is only possible if you also set the option `register` to `true`. This may be a leftover from old versions.

### Versions

`@axa-fr/oidc-client@7.22.22`

#### Expected

Setting the `service_worker_register` option should be enough.

#### Actual

You have to set the option `register` to a truly value.

### Additional Details

The affected section contains a query that is probably outdated:

```js
  if (configuration.register) {
    registration = await configuration.service_worker_register(serviceWorkerRelativeUrl);
  } else {
    registration = await navigator.serviceWorker.register(serviceWorkerRelativeUrl);
  }
```

I tried to maintain backwards compatibility with my fix:

```js
  if (configuration.register || configuration.service_worker_register) {
    registration = await configuration.service_worker_register(serviceWorkerRelativeUrl);
  } else {
    registration = await navigator.serviceWorker.register(serviceWorkerRelativeUrl);
  }
```

guillaume-chervet:
this issue sould be ok now @thiscode thank you very much for your help !

PR #1440, opened 2024-08-23
Fix usage of config option service_worker_register

thiscode:
This will fix the Issue #1439

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -106,7 +106,7 @@ export const initWorkerAsync = async (configuration, configurationName) => {
   }
 
   let registration = null;
-  if (configuration.register) {
+  if (configuration.register || configuration.service_worker_register) {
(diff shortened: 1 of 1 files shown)


packages/oidc-client/src/initWorker.ts
```diff
@@ -106,7 +106,7 @@ export const initWorkerAsync = async (configuration, configurationName) => {
   }
 
   let registration = null;
-  if (configuration.register) {
+  if (configuration.register || configuration.service_worker_register) {
```
guillaume-chervet, line 109: thank you so much @thiscode 
It should be only
if (configuration.service_worker_register) {

I have checkout the code and configuration.register does not exist any more. I should add more Typing to this file.

guillaume-chervet: I squash merge your PullRequest, Thank you @thiscode ! it is already better.

PR #1448, opened 2024-09-05
fix(oidc): ensure unique tabIds when page is duplicated (release)

krzempekk:
This is a follow-up for https://github.com/AxaFrance/oidc-client/pull/1423. Mechanism introduced in https://github.com/AxaFrance/oidc-client/pull/1402 works under assumption that session storage for each tab is independent and opening new tab creates fresh instance of it. Which is generally true, except for one browser gimmick - if page is duplicated (in Chrome -> right click on tab name then "Duplicate") session storage is copied to duplicated tab (see https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). This creates an issue in following scenario:
1. User opens application and logs in
2. User duplicates tab 
3. User logs out from both tabs
4. User tries to log in again on both tabs
This scenario can cause various issues, depending on specific IDP, which are caused by two tabs having the same id. 

## After this PR

This PR introduces solution inspired by https://github.com/haricane8133/unique-tabid/tree/master, which is based on BroadcastChannel. Idea is that each `Oidc` class instance has channel listening for messages. When new instance is created (either in duplicated tab, if app url was opened in empty tab or if page was refreshed), following steps are done:
1. BroadcastChannel instance is created
2. Session storage is checked
     - If ID is not present, then new ID is generated and logic ends
     - If ID is present then `SEARCH` message is sent via channel
3. Other tabs, if there are any, are receiving `SEARCH` message and comparing it with their own ID. If there is a match, `FOUND` message is sent.
4. If `FOUND` message is received, new ID is generated. Otherwise, if no message is received after some timeout (assumed 500ms here), ID from session storage is used.

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -57,17 +57,8 @@ export const defaultServiceWorkerUpdateRequireCallback =
     location.reload();
   };
 
-const getTabId = (configurationName: string) => {
-  const tabId = sessionStorage.getItem(`oidc.tabId.${configurationName}`);
-
-  if (tabId) {
-    return tabId;
-  }
-
-  const newTabId = globalThis.crypto.randomUUID();
-  sessionStorage.setItem(`oidc.tabId.${configurationName}`, newTabId);
-  return newTabId;
-};
+const getTabId = (configurationName: string) =>
+  sessionStorage.getItem(`oidc.tabId.${configurationName}`);
 
 const sendMessageAsync =
   registration =>
--- a/packages/oidc-client/src/keepSession.ts
+++ b/packages/oidc-client/src/keepSession.ts
@@ -17,6 +17,7 @@ export const tryKeepSessionAsync = async (oidc: Oidc) => {
       configuration.authority,
       configuration.authority_configuration,
     );
+    await oidc.ensureUniqueTabId();
     serviceWorker = await initWorkerAsync(configuration, oidc.configurationName);
     if (serviceWorker) {
       const { tokens } = await serviceWorker.initAsync(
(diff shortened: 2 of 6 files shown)


krzempekk:
Hi @guillaume-chervet ! Could you take a look at this?

guillaume-chervet:
Hi @krzempekk , I test it as soon as possible (like today) Thank you again. I did not know that tab clon can bring some problems. Thank you!

guillaume-chervet:
I do not know how but it would be awesome to move the main part of the code in initWorker.js .

krzempekk:
Hi @guillaume-chervet! I tried to put this code in `initWorker.js`, but I haven't found reasonable way to do this. Adding `tabId` and `channel` to `Oidc` class helps ensuring that there is only one instance of them per tab.

guillaume-chervet:
Let ='s try it as a beta.
I Will try to find an idea but let's iterate.

Thank you @krzempekk :) again !

krzempekk:
Hi @guillaume-chervet ! Thanks for merging! Can you trigger release with this fix? I see that it was not triggered

guillaume-chervet:
It was published yesterdays as a beta. I prefer to test it well before release it!

krzempekk:
Sure, totally understand! Could you share beta version so I can test it as well?

guillaume-chervet:
Version 7.22.25-beta.1555 is already on npm @krzempekk

grpawel:
Hi @guillaume-chervet, I noticed this PR didn't end up in any released version, looks like it was removed from `main` branch.
Could you shed some light if you've discovered some issues in beta version? We would like to fix the solution so it's good enough to be published in a patch release.

Issue #1400, opened 2024-07-05
Loop of Service Worker initialisation when local app opened and reopened with different oidc package version installed


piotrbartnik:
### Issue and Steps to Reproduce

When we change the version of `oidc-client` and `oidc-client-service-worker` in local development between runs sometimes the app can't be loaded properly again.

App is reloaded couple of times, sometimes infinitely.

It seems that service worker for some reason is not activated properly on downgraded/upgraded version.

It is triggered inside `initWorkerAsync.ts`.
When `if(serviceWorkerVersion !== codeVersion)` the console log about unregistering and reloading is prompted.
Then the `service_worker_update_require_callback` kicks in with default reload of page.

It happens in my personal react app where I try similar steps but also for `AxaFrance/oidc-client.git` demo app with below steps.
Sometimes the `index.js:255 Service worker 7.22.9 version mismatch with js client version 7.22.8, unregistering and reloading` happens once, sometimes twice, sometimes multiple times (together with reloading, unregistering and registering service worker again - basically whole process).

* clone and install `git clone https://github.com/AxaFrance/oidc-client.git`,
* uncomment service worker mode (both `service_worker_relative_url: '/OidcServiceWorker.js',` in `examples/react-oidc-demo/src/configurations.ts`),
* run the react app locally with `pnpm start`,
* close dev server,
* change oidc versions in `packages/oidc-client/src/version.ts` and `packages/oidc-client-service-worker/src/version.ts` to some lower version (in my case to `7.22.8` ,
* install packages again,
* run the dev server again,
* see the console

When the problem with more than one reload occurs the console logs seem to follow the pattern:
```
> Service worker 7.22.9 version mismatch with js client version 7.22.8, unregistering and reloading
> [OidcServiceWorker] service worker installed 1720185015
> TypeError: Failed to fetch
    at me ()
    at r ()
    at Object.startKeepAliveServiceWorker ()
    at mn ()
> [vite] connecting...
> [vite] connected.
>  Service worker 7.22.9 version mismatch with js client version 7.22.8, unregistering and reloading
... looped again
```

### Versions
Tried with 7.22.9 -> 7.22.8 -> 7.22.9
as well as with 7.22.9 -> 7.22.7 -> 7.22.9 
Browser: chrome latest (haven't tested on other browsers),

### Screenshots
Axa fr demo react app single reload (proper(?) behavior):
[...]

PR #1449, opened 2024-09-13
fix(oidc): force sw update when new registration appears (release)

wermanoid:
Should help with https://github.com/AxaFrance/oidc-client/issues/1400

Also allow to update SW when some code changes happens within same version. At lest existing `service_worker_update_require_callback` was effective enough for me in private app tests and with react-oidc-demo app too. I didn't fail in any reload loop, since SW copy function was added to app prestart.

@piotrbartnik  @guillaume-chervet  would be great if you could prove that it works or not at yours environment

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -113,6 +113,11 @@ export const initWorkerAsync = async (
     registration = await configuration.service_worker_register(serviceWorkerRelativeUrl);
   } else {
     registration = await navigator.serviceWorker.register(serviceWorkerRelativeUrl);
+
+    if (registration.active && registration.waiting) {
+      console.log('Detected new service worker waiting, unregistering and reloading');
+      await configuration.service_worker_update_require_callback?.(registration, stopKeepAlive);
+    }
   }
 
   try {
@@ -151,7 +156,7 @@ export const initWorkerAsync = async (
       console.warn(
         `Service worker ${serviceWorkerVersion} version mismatch with js client version ${codeVersion}, unregistering and reloading`,
       );
-      await oidcConfiguration.service_worker_update_require_callback(registration, stopKeepAlive);
+      await oidcConfiguration.service_worker_update_require_callback?.(registration, stopKeepAlive);
     }
 
     // @ts-ignore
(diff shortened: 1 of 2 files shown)


packages/oidc-client/src/initWorker.ts
```diff
@@ -113,6 +113,11 @@ export const initWorkerAsync = async (
     registration = await configuration.service_worker_register(serviceWorkerRelativeUrl);
   } else {
     registration = await navigator.serviceWorker.register(serviceWorkerRelativeUrl);
+
+    if (registration.active && registration.waiting) {
+      console.log('Detected new service worker waiting, unregistering and reloading');
+      await configuration.service_worker_update_require_callback?.(registration, stopKeepAlive);
```
guillaume-chervet, line 119: hi  @wermanoid thank you for your PullRequest !
I'am not sure to understand why we nee 2 time  await configuration.service_worker_update_require_callback?.(registration, stopKeepAlive); ? may you help me to understand?

wermanoid: there are cases when new service worker is added, when old one is still running. This leads to case when SW is just hanging in `waitting` state and never starts, as original SW is still running as usual.

As far as I understand now, oidc is able to update SW only when running SW version and oidc version mismatch.
But we had cases when ui was redeployed and that leads to case, when 2 SW of same internal oidc version appeared in browser.

guillaume-chervet: Thanh you @wermanoid I will test your version tomorrow! sorry for the delay.

wermanoid: Hi @guillaume-chervet , any chance to verify? Or can I help somehow to move forward?

piotrbartnik: I checked it referring to my Issue [#1400](https://github.com/AxaFrance/oidc-client/issues/1400)

I can't reproduce the sw mismatched versions issue anymore in demo app running from [wermanoid:fix/update-sw-registration](https://github.com/wermanoid/oidc-client/tree/fix/update-sw-registration)

guillaume-chervet (approved):

guillaume-chervet:
very very sorry for the delay @wermanoid

Issue #1481, opened 2024-11-06
Missing changes from #1448


grpawel:
Hi, this is not an issue, I'm just not sure if discussion in already closed PRs are monitored.
A couple months ago there was a PR #1448 published as 7.22.25-beta.1555. Although GitHub shows it was merged to main branch, the changes and commits are actually missing in this repo. Was the commit removed with force-push maybe?

Did you discover any issues with the solution during testing? We'd like to open a new PR with the same changes if it works correctly.

guillaume-chervet:
Hi @grpawel , It was freezing all my browsers. I had to remove the change.
I have to invistigate why. 
It would be greatd to re-open the pullRequest to continue to work on it.

PR #1485, opened 2024-11-18
fix(oidc): ensure unique tabIds when page is duplicated (beta)

grpawel:

This PR has the same changes as #1448, which was merged but then removed (see discussion in #1481).

See full description at #1448.

## Before this PR
When user duplicates a browser tab, session storage is copied so both tabs use the same `tabId`. Operations like login in one tab can override the state of another tab.

## After this PR
OIDC client can detect that the tab was duplicated and its `tabId` is already used.

### DIFF
--- a/packages/oidc-client/src/initWorker.ts
+++ b/packages/oidc-client/src/initWorker.ts
@@ -57,17 +57,8 @@ export const defaultServiceWorkerUpdateRequireCallback =
     location.reload();
   };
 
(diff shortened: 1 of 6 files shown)


guillaume-chervet:
Very sorry for the delay @grpawel .

guillaume-chervet:
Hi @grpawel , very sorry for the conflict. May you resolved it and I have time tomorrow to test it well.

PR #1603, opened 2025-09-10
Clear userInfo in storage in clearAsync function

mteterel:
When using localStorage as the session storage, the userInfo stays present even after logging out. Due to this and with the recent addition of caching, invalid data gets eventually loaded from the localStorage (as there is no validation on load) and will display wrong information in some cases (e.g. user has switched account on the IdP).


guillaume-chervet:
Thank you so much @mteterel, sorry for the delay, it test for it tomorrow :)

guillaume-chervet (approved):

PR #1606, opened 2025-09-16
feat(react-oidc): enforce uniqueness of redirect_uri and silent_redirect_uri

u0xaa55:
Raise an exception in case `redirect_uri` is the same as `silent_redirect_uri`
image

### DIFF
--- a/packages/react-oidc/README.md
+++ b/packages/react-oidc/README.md
@@ -182,6 +182,9 @@ const App = () => (
 render(<App />, document.getElementById('root'));
 ```
 
+> [!WARNING]
+> If you have both `redirect_uri` and `silent_redirect_uri` configured, their value must be different.
+
 ```javascript
 const configuration = {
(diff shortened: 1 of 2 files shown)


u0xaa55:
> Awesome @u0xaa55 ! Thank you !
> 
> it can be better I think to set up this condition in the vanilla js part here =>
> 
> https://github.com/AxaFrance/oidc-client/blob/c781911387a704979bba4a084670e507473325aa/packages/oidc-client/src/oidc.ts#L119
> 
> What do you think?

since the reason why this is not supported [lies in `OidcRoutes`](https://github.com/AxaFrance/oidc-client/blob/c781911387a704979bba4a084670e507473325aa/packages/react-oidc/src/core/routes/OidcRoutes.tsx#L46) and `OidcRoutes` is used by `react-oidc` only (not oidc-client), I thought there's no reason to block people using `oidc-client` directly. Let me know if you have a better approach :)

Issue #1274, opened 2024-02-01
"undefined" is not valid JSON while authenticating once deployed, working fine in local.


bhargavmullakuru:
### Issue and Steps to Reproduce
Hi @guillaume-chervet , App was working fine in local, After deploying to server, it was throwing the below error.

"undefined" is not valid JSON at `getLoginParams`, Could you pls help me with this.
Her is my config format. 

```
{
authority:''
client_id:''
redirect_uri:''
response_type:''
scope:''
 storage: localStorage,
demonstrating_proof_of_possession : true
demonstrating_proof_of_possession_configuration : config
}
```

### Versions
7.14.1
### Screenshots
![Screenshot 2024-02-01 214606](https://github.com/AxaFrance/oidc-client/assets/26109320/e9c320d5-e54d-4bd3-88d3-b1d91db0447d)
![Screenshot 2024-02-01 214757](https://github.com/AxaFrance/oidc-client/assets/26109320/79f39326-19fd-4c07-8993-da63f23bbca0)

### Additional Details

- Installed packages:

bhargavmullakuru:
@guillaume-chervet,  Seems like storage is not working when deployed in server. When debugged, the same storage object in local has diff values, whereas server Storage object doesn't have any values.

In local.
![image](https://github.com/AxaFrance/oidc-client/assets/26109320/c057875e-5e8d-4f2f-9803-f62a9cac8fd4)

in Dev, Storage was not created properly
![Screenshot 2024-02-01 224133](https://github.com/AxaFrance/oidc-client/assets/26109320/ea219fb9-50d4-4030-8856-e8e76dc1b1a2)

guillaume-chervet:
hi @bhargavmullakuru , do you have a video which can help me to unsderstand user flow on your server?

At which stage does this happen?

bhargavmullakuru:
@guillaume-chervet  Just normally deployed and tried to access Dev URL, it is throwing that error.
Let me do and send a screen recording

guillaume-chervet:
hi @bhargavmullakuru ,

You may have a wrong configuration somewhere. I need more information to be able help you.

bhargavmullakuru:
Yes @guillaume-chervet , but it's working fine in local. 🫣🥲

Moraissz:
@bhargavmullakuru  We have the same problem, but even local it doesn't work

We figure out that its because the redirectUri param did not finish with '/authentication/callback'.

[...]

Issue #1257, opened 2024-01-05
Uncaught SyntaxError: "undefined" is not valid JSON


gitawego:
### Issue and Steps to Reproduce

when calling method `getLoginParams` in file `initSession.ts`, if there is no param set in localStorage, then it throws error 

```
Uncaught SyntaxError: "undefined" is not valid JSON
```

It happens very frequently. 

based on the code, `JSON.parse` can't parse the `undefined` value, if `dataString` is undefined, then oidc can't initialize the session.

```ts
const getLoginParams = () => {
        const dataString = storage[`oidc.login.${configurationName}`];
        if (!getLoginParamsCache[configurationName]) {
            getLoginParamsCache[configurationName] = JSON.parse(dataString);
        }
        return getLoginParamsCache[configurationName];
    };
```

maybe we should add a default value to dataString, like

```ts
const dataString = storage[`oidc.login.${configurationName}`] || 'null';
```

### Versions

all

guillaume-chervet:
Hi @gitawego , 
Thank you for your issue.
Which version of the library are you using ?
Do you have a sample (ofuscated) of your configuration ?

I'am searching from a long time why storage can be null. It never set to null in the code and values are required to refresh tokens correctly.

gitawego:
I put `null` is only to make JSON.parse works, because undefined can't be parsed. 

in this line of code: https://github.com/AxaFrance/oidc-client/blob/main/packages/oidc-client/src/initSession.ts#L66
```js
const dataString = storage[`oidc.login.${configurationName}`];
```

`dataString` could be `undefined`,

and when parsed at line 68: https://github.com/AxaFrance/oidc-client/blob/main/packages/oidc-client/src/initSession.ts#L68

```js
getLoginParamsCache[configurationName] = JSON.parse(dataString);
```

if `dataString` is undefined, it throws error, because `JSON.parse` can't parse undefined.


```js
JSON.parse(JSON.stringify(undefined)) // throw Uncaught SyntaxError: "undefined" is not valid JSON
JSON.parse(undefined) //Uncaught SyntaxError: "undefined" is not valid JSON
JSON.parse(null) // return value null
JSON.parse(JSON.stringify(null)) // return value null
```

in my company, it's v7.6.0, but I compared, this part of the code is never changed.

guillaume-chervet:
I will manage this case with a warning instead of a crash.
It would not be a problem for 90% of configurations cases.

[...]

Issue #871, opened 2022-09-27
undefined is not valid JSON


beanovar:
Hi - I'm sorry at this point I'm not sure what it takes to reproduce, will post if I find out. 
However sometimes the page errors out with the following logs in the console: 

```
SyntaxError: "undefined" is not valid JSON
   at JSON.parse (<anonymous>)
   at getLoginParams (oidc.ts:199:1)
   at Oidc._callee23$ (oidc.ts:963:1)
   ....
   at localsilentLoginAsync (oidc.ts:961:1)
refreshTokensAsync_silent_error {"message":"exceptionSilent","exception":"\"undefined\" is not valid JSON"}
```

Do you know what would cause this at all? 

Thankyou

guillaume-chervet:
Hi @beanovar thank you very much for your issue.

Do you have a sample of your configuration?
Does it occur quickly or after a big amount of time?
Does it occur often?

beanovar:
Hi - I might see it every 3 or so days? Only after a big amount of time, maybe while the page is in the background. Haven't seen the pattern yet, or how to reproduce. 

Our config sets 
- client_id
- redirect_uri
- silent_redirect_uri
- silent_login_uri
- scope including offline_access
- authority
- authority_time_cache_wellknowurl_in_second
- refresh_time_before_tokens_expiration_in_second
- service_worker_relative_url
- service_worker_only -> true
- monitor_session -> false

I will let you know if I see what can re-produce it.
Thanks

guillaume-chervet:
Thank you @beanovar, on which browser do you have this error?

beanovar:
Chrome - I'm going to close this for now as I have not seen it for a while and think it was to something to do with the authentication server returning bad responses. It might be helpful to review what the lib might do if the auth server returns unexpected results or errors? It could be something else entirely though, just guessing. Thanks!

guillaume-chervet:
Thank you @beanovar , i think it migth have something link with session livecycle. If you have the problem back you can reopen the issue.

beanovar:
I found out how to reproduce this ! 
1) Have two tabs open, one in the foreground, one in the background (hidden)
2) Ensure the auth server is down / returns an error response when the visible tab refreshes it's token.
3) Let it retry the 4 times before eventually failing. 
4) Now make the second tab visible, and you will see the "undefined" is not valid JSON in that tab's console.

[...]

PR #1674, opened 2026-06-09
fix(oidc): prevent storage poisoning on undefined setters (alpha)

pixelmord:

```js
// in a browser console:
localStorage['oidc.login.default'] = undefined;
localStorage.getItem('oidc.login.default');
// → "undefined"   (the literal string)

JSON.parse(localStorage.getItem('oidc.login.default'));
// → Uncaught SyntaxError: "undefined" is not valid JSON
```

## Before this PR

Every setter in `initSession.ts` (and the SW fallback path in `initWorker.ts`) writes through bracket notation:

```ts
loginStorage[`oidc.login.${configurationName}`] = JSON.stringify(data);
```

When `data === undefined`, `JSON.stringify(undefined)` returns the JS value `undefined`. The Web Storage spec then coerces that to the literal string `\"undefined\"` on bracket-notation assignment. The next read of `getLoginParams` calls `JSON.parse(\"undefined\")` → throws.

We have observed this throw **3593 times across 391 users in 6 months** in production logs.

Prior reports/PRs for the same symptom:
- #871 (2022-09) — closed, no fix
- #1257 (2024-01) — closed after PR #1258 added a getter guard
- #1274 (2024-02) — closed as duplicate

PR #1258 added `if (!dataString)` in `getLoginParams`, but `\"undefined\"` is a *truthy* string so that guard misses the poisoning case. The setter side has never been addressed, and the same pattern exists for other keys (`oidc.${name}` tokens, `oidc.jwk.${name}`, `oidc.nonce.${name}`, `oidc.state.${name}`, `oidc.code_verifier.${name}`, `oidc.session_state.${name}`).

## After this PR

1. **Every setter is guarded.** If the input is `undefined`/`null`, the key is deleted instead of being assigned a coerced string. This covers `setLoginParams`, `setTokens`, `setNonceAsync`, `setStateAsync`, `setCodeVerifierAsync`, `setSessionStateAsync`, and both DPoP setters.

2. **Defence in depth on the read side.** A new `parseJsonOrNull` helper treats the literal strings `\"undefined\"`, `\"null\"`, and `\"\"` as missing, so storage already poisoned by older versions of the library self-heals on the next read (`initAsync`, `getTokens`, `getLoginParams`, `getDemonstratingProofOfPossessionJwkAsync`).

3. **SW fallback path in `initWorker.ts`** gets the same treatment for `setLoginParams`/`getLoginParams` (it has its own copy of this code outside `initSession`).

4. **Regression tests** in `initSession.spec.ts` cover both new poisoning attempts (calling setters with `undefined`/`null`) and recovery from pre-existing poisoned storage entries written by older builds.

## Verification recipe

1. `localStorage['oidc.login.default'] = undefined` (in DevTools).
2. `localStorage.getItem('oidc.login.default')` → returns the string `\"undefined\"`.
3. Trigger any silent refresh / login callback path → reproducible throw on `main`.
4. With this PR applied, step 1 deletes the key (or, if storage is pre-poisoned, the next read self-heals).

## Test plan

- [x] `pnpm exec vitest run --root .` in `packages/oidc-client` → 63 / 63 pass (9 new).
- [x] `pnpm exec tsc --noEmit` clean.
- [x] `pnpm exec prettier --check` and `pnpm exec eslint` clean.

## Commit tag

Tagged `(alpha)` per `CONTRIBUTING.md` — happy to retag to `(beta)` or `(release)` if you'd prefer given the production impact (the fix is intentionally narrow and additive-only, no API or behaviour change for any non-poisoned input).

### DIFF
--- a/packages/oidc-client/src/initSession.spec.ts
+++ b/packages/oidc-client/src/initSession.spec.ts
@@ -165,6 +165,71 @@ describe('initSession', () => {
     });
   });
 
+  describe('undefined/null guards (regression: #871 / #1257 / #1274)', () => {
+    let storage: Storage;
+    let session: ReturnType<typeof initSession>;
+
+    beforeEach(() => {
+      storage = makeStorage();
+      session = initSession(configName, storage);
+    });
+
+    it('setLoginParams(undefined) does not poison storage with the literal "undefined"', () => {
+      session.setLoginParams(undefined);
+      expect(storage[`oidc.login.${configName}`]).toBeUndefined();
+      expect(() => session.getLoginParams()).not.toThrow();
+      expect(session.getLoginParams()).toBeNull();
+    });
+
+    it('setLoginParams(null) clears storage rather than writing "null"', () => {
+      session.setLoginParams({ callbackPath: '/cb', extras: null, scope: 'openid' });
+      session.setLoginParams(null);
+      expect(storage[`oidc.login.${configName}`]).toBeUndefined();
+      expect(session.getLoginParams()).toBeNull();
+    });
+
+    it('getLoginParams tolerates a pre-existing poisoned "undefined" value', () => {
+      // simulate storage poisoned by an older version of the library
+      storage[`oidc.login.${configName}`] = 'undefined';
+      expect(() => session.getLoginParams()).not.toThrow();
+      expect(session.getLoginParams()).toBeNull();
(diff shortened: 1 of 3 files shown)


guillaume-chervet:
Thank you so much @pixelmord

PR #1720, opened 2026-07-28
feat(oidc-client): add typed-error

guillaume-chervet:


guillaume-chervet:
@copilot can you review the architecture of the modifications. Does it breack existing interface? Could it be better?

guillaume-chervet:
@copilot, are you able to enhance?  The main compatibility risk is in the silent-login iframe path: deserializeOidcError() always recreates a plain OidcError, so an OidcStateError thrown in the iframe comes back as OidcError on the parent side. That means isOidcStateError(error) / instanceof OidcStateError can fail in silent renew, even though the code is still correct. There is also an inconsistency where NONCE_MISSING is an OidcStateError in login.ts but a plain OidcError in renewTokens.ts.

So: no broad interface break for regular Error consumers, but there is a subtype-identity compatibility hole for callers relying on isOidcStateError. I think the architecture is good in principle, but it would be better either to reconstruct the specialized subclasses during deserialization, or to standardize on code-based checks only and avoid relying on instanceof acros

guillaume-chervet:
@copilot does actual pullrequest do a breaking change?

guillaume-chervet:
@copilot did the pullrequest make breacking change? Can the resolution can be enhanced?

guillaume-chervet:
You should have keep this previous little breaking change @copilot , I lile it. Set it back again

PR #1725, opened 2026-09-12
fix(oidc): ignore bare trailing hash in getPath

### LINKED ISSUE #1724: Bare trailing # is included in getPath() and prevents callback route matching

`getPath()` treats a trailing empty fragment (`#`) as part of the route.

This can be reproduced directly with:

```ts
getPath(
  "https://example.com/authentication/callback?state=abc&code=def#",
);
```

Current result:

```text
/authentication/callback#
```

Expected result:

```text
/authentication/callback
```

The current implementation already normalizes `#_=_` to an empty fragment, but a bare `#` is left intact and appended to the pathname.

This matters in `@axa-fr/react-oidc`, where `OidcRoutes` uses `getPath()` to match the current callback route against the configured redirect URI. With a callback URL ending in `#`, the comparison becomes:

```text
/authentication/callback#
!=
/authentication/callback
```

so the callback route is not recognized and the callback component is not rendered.

### Versions

- Observed with `@axa-fr/react-oidc` 7.29.1
- The same `getPath()` behavior is still present on the current `main` branch

### Expected

A fragment delimiter with no fragment content should not alter the route returned by `getPath()`.

```text
/authentication/callback
```

### Actual

The bare fragment delimiter is appended to the returned path:

```text
/authentication/callback#
```

This can prevent the OIDC callback route from matching.

### Additional Details

Non-empty hashes should continue to be preserved for hash-routing support.

This issue is specifically about an empty fragment delimiter (`#`), not hash fragments in general.

### Screenshots

N/A — the issue is reproducible directly at the `getPath()` utility level.

m4rtinpf:
Fixes #1724

## Before

`getPath()` appended a bare trailing `#` to the returned route.

For example:

```ts
getPath(
  "https://example.com/authentication/callback?state=abc&code=def#",
);
```

returned:

```text
/authentication/callback#
```

In `@axa-fr/react-oidc`, this could prevent `OidcRoutes` from matching the OIDC callback route.

## After

A bare trailing `#` is treated as an empty fragment:

```text
/authentication/callback
```

Non-empty hash fragments keep their existing behavior.

A regression test has been added for the empty-fragment case.


guillaume-chervet (commented):
Thank you @m4rtinpf  !

guillaume-chervet (approved):

